PR #1, opened 2016-02-16
imap-hibernate: send connect/disconnect events to anvil

manuelm:
This makes `doveadm who` and `mail_max_userip_connections` work with hibernated connections.


sirainen:
Merged.

PR #2, opened 2016-02-23
Is there any document  about how to dovelop a new plugin

kevinsir:
I am  a newer! I want to implement a function  when a new mail is coming to dovecot , I can execute my custom program,something like insert mail to database!
I find a notify plugin in dovecot.
How can I use it? I have makefile the sourcecode of the dovecot, and compile it and also see the code of mail-log , notify! 
Is there any document  about how to dovelop a new plugin, or  is there any other easy way?
thanks!


cmouse:
Please send email to dovecot@dovecot.org for questions, thank you!

PR #3, opened 2016-03-22
Allow multiple recipient delimiter

m42e:
This will allow to use multiple different recipient delimiter, e.g. `+` and `-`


aclindsa:
Though I'm not an expert on this code, I think https://github.com/dovecot/core/pull/4 may be a more complete patch (authored by @lhw). Do you agree, or do you not think all the additional changes are necessary?

m42e:
@aclindsa Well You are right, because it fixes the address for further processing (e.g. calling lmtp_proxy_add_rcpt). So maybe this seems to be a bit hacked :) and may not work under every condition :(

PR #4, opened 2016-03-22
Support multiple recipient_delimiters

aclindsa:
The recipient_delimiter is treated as multiple one-character delimiters
rather than one multi-character delimiter if more than one character is
supplied. The address is split on the first character in
recipient_delimiter found.

### DIFF
--- a/src/lmtp/commands.c
+++ b/src/lmtp/commands.c
@@ -292,11 +292,11 @@ client_proxy_is_ourself(const struct client *client,
 }
 
 static const char *
-address_add_detail(struct client *client, const char *username,
+address_add_detail(const char *username, const char delim_c,
 		   const char *detail)
 {
-	const char *delim = client->unexpanded_lda_set->recipient_delimiter;
 	const char *domain;
+	const char delim[] = {delim_c, '\0'};
 
 	domain = strchr(username, '@');
 	if (domain == NULL)
@@ -308,7 +308,7 @@ address_add_detail(struct client *client, const char *username,
 }
 
 static bool client_proxy_rcpt(struct client *client, const char *address,
-			      const char *username, const char *detail,
+			      const char *username, const char *detail, char delim,
 			      const struct lmtp_recipient_params *params)
 {
 	struct auth_master_connection *auth_conn;
@@ -363,7 +363,7 @@ static bool client_proxy_rcpt(struct client *client, const char *address,
 		if (*detail == '\0')
 			address = username;
 		else
-			address = address_add_detail(client, username, detail);
+			address = address_add_detail(username, delim, detail);
 	} else if (client_proxy_is_ourself(client, &set)) {
 		i_error("Proxying to <%s> loops to itself", username);
 		client_send_line(client, "554 5.4.6 <%s> "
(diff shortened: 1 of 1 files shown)


aclindsa:
@sirainen Would you mind taking a look at this patch? I'm glad to revise it and address feedback if needed.

sirainen:
The only minor thing is that for coding style consistency just use "char delim" instead of "const char delim" in the parameters. Anyway this changes existing behavior (it was already possible to use delimiter-string), so it's a v2.3 change.

aclindsa:
@sirainen Thanks for taking a look! Pardon my ignorance about the dovecot development cycle, but what do I need to do to ensure this makes it into v2.3 (and that the accompanying pigeonhole patch goes in at the same time)?

Also, I think I've addressed your 'const' comment, but let me know if you meant something different from the current patch.

src/lmtp/commands.c
```diff
@@ -308,7 +308,7 @@ address_add_detail(struct client *client, const char *username,
 }
 
 static bool client_proxy_rcpt(struct client *client, const char *address,
-			      const char *username, const char *detail,
+			      const char *username, const char *detail, char delim,
```
cmouse: this should be const char delim

aclindsa: This originally was const, but I changed it after the earlier comment by @sirainen. I'm willing to change it back if that is the consensus, or if I misinterpreted his earlier comment.

cmouse:
Ah my bad.

cmouse:
We are currently reviewing your patch internally.

aclindsa:
Closing since this has been committed as https://github.com/dovecot/core/commit/972c9172e9e6a0fc6053efb3d2ee9d354b67727f

jnbt:
Are there any plans for this patch to be included in a future release version?

cmouse:
As you can see it has been merged, outside github though.

jnbt:
Thank for your fast reaction time. I understood that this patch was merged as 972c917 into master, but 972c917 is missing from the release branches, e.g. version 2.2.28: https://github.com/dovecot/core/commits/2.2.28/src/lmtp/commands.c

I would assume this commit (972c9172e9e6a0fc6053efb3d2ee9d354b67727f) to be between aeea3dbd1f4031634f7b318614adf51dcfc79f42 and 6de95f63d4590814354fa5e35e92946c58d4562e.

cmouse:
it will probably be in 2.3

jnbt:
OK thanks!

matrixise:
Some news about this commit? and when you think to release 2.3? 

Thank you

cmouse:
It's already in v2.3 (which is currently master).

matrixise:
@cmouse I wanted to tell, a real tarball ;-) yep, I have seen the commit in master, thank you.

in fact, I use the tvial/docker-mailserver docker image, and I need the support of a multi recipient delimiter because sometimes some sites do not support the '+' character :/ and the alternative is '-'. Issue, I already have some emails with the '+' character. 

Do you have a milestone ?

Thank you

PR #5, opened 2016-03-29
fts-squat: fix Corrupted squat uidlist bug

nikwrt:
Hello,

I have experienced the "Corrupted squat uidlist" bug described in these two threads:

http://dovecot.org/list/dovecot/2012-April/135162.html
http://dovecot.org/list/dovecot/2016-January/102864.html

and I have tested a patch based on the explanation in the second thread. This has been working flawlessly for approximately a month. I'd really like to see this merged so I don't have to repatch it every time.

### DIFF
--- a/src/plugins/fts-squat/fts-backend-squat.c
+++ b/src/plugins/fts-squat/fts-backend-squat.c
@@ -96,7 +96,7 @@ static void fts_backend_squat_deinit(struct fts_backend *_backend)
 	i_free(backend);
 }
 
-static void
+static int
 fts_backend_squat_set_box(struct squat_fts_backend *backend,
 			  struct mailbox *box)
 {
@@ -105,12 +105,22 @@ fts_backend_squat_set_box(struct squat_fts_backend *backend,
 	struct mailbox_status status;
 	const char *path;
 	enum squat_index_flags flags = 0;
+        int ret;
 
 	if (backend->box == box)
-		return;
+        {
+		if (backend->refresh) {
+                        ret = squat_trie_refresh(backend->trie);
+                        if (ret < 0)
+				return ret;
+			backend->refresh = FALSE;
+		}
+		return 0;
+	}
 	fts_backend_squat_unset_box(backend);
+	backend->refresh = FALSE;
 	if (box == NULL)
-		return;
+		return 0;
 
 	perm = mailbox_get_permissions(box);
 	storage = mailbox_get_storage(box);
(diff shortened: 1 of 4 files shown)


nikwrt:
Hi, I have been running this now for approximately six months, with no issues whatsoever. Please, consider merging it.

cmouse:
I'll try to have a look at this next week. Sorry for the delay.

cmouse:
We'll do internal review for this

src/plugins/fts-squat/fts-backend-squat.c
```diff
@@ -137,6 +147,9 @@ fts_backend_squat_set_box(struct squat_fts_backend *backend,
 	if (backend->full_len != 0)
 		squat_trie_set_full_len(backend->trie, backend->full_len);
 	backend->box = box;
+	ret = squat_trie_open(backend->trie);
+	if (ret < 0)
+		return -1;
```
sirainen: This is missing return value for ret >= 0 case.

nikwrt: This is intended. If ret >= 0 the next line will return what the original code did:

return squat_trie_get_last_uid(backend->trie, last_uid_r);

Previously, squat_trie_open could fail and squat_trie_get_last_uid would still be called, originating the error.

Hope this helps...

sirainen: But you are missing return value from the function entirely when ret >= 0, so it'll now more or less randomly return something. I'd guess the above should be: return squat_trie_open(backend->trie);

nikwrt: I am puzzled, are we looking at the same code? Please click maybe on "View full changes". If ret >=0, the next line will execute, and that is:

`return squat_trie_get_last_uid(backend->trie, last_uid_r);`

sirainen: You're looking at fts_backend_squat_get_last_uid(), I'm looking at fts_backend_squat_set_box().

nikwrt: Ah yes, I see it now. You are right. It should return 0 and I overlooked it.

nikwrt: Just wondering, should I fix it in my fork and resubmit it? Just let me know.

sirainen: Yeah, that'd be easier for us :)

nikwrt: It's done now and I see it's been picked up automatically by the pull request. Thanks for pointing it out.

sirainen:
merged

PR #6, opened 2016-04-09
lmtp: Use port 24 if no port has been provided

wido:
This allows using the LMTP, IMAP and POP3 proxy on the same
Dovecot installation without the userdb providing the port
to connect to.

TCP port 24 is registered at IANA as: "any private mail system"

LMTP being the Local Mail Transfer Protocol classifies as a private
mail system and thus justifies the usage of port 24.

Prior to this patch the LTMP client would connect to TCP port 0 by
default if the userdb did not provide a port to connect to.


wido:
@cmouse Could you take a look at this one?

cmouse:
Sorry for huge delay, I'll take a look.

wido:
Thank you @cmouse 

Any comments on this one?

cmouse:
Hi!

We will consider this for v2.3.

cmouse:
Hi, we'll do internal review on this

cmouse:
Merged.

PR #7, opened 2016-04-09
quota_clone plugin: quota_clone_mail_user_created warned without debug

bdraco:
"The quota_clone_dict setting is missing from configuration" warning
would be thrown when debug mode was not enabled.  This warning
happens under normal circumstances when prefix="" namespace is not
defined (usually when set to INBOX.)

See http://dovecot.org/list/dovecot-cvs/2012-October/021981.html for
more details on the automagic prefix="" namespace.

This change suppresses the spurious warning.


cmouse:
We'll perform internal review on this.

cmouse:
Merged.

PR #8, opened 2016-04-15
check for NULL

Laeeth:
managesieve was complaining about not knowing virtual driver for virtual namespace, which meant process kept dying (due to error in lib-storage).  as a quick kludge I set the following flag in main.c for managesieve.
 MAIL_STORAGE_SERVICE_FLAG_NO_NAMESPACES

However that leads to mail-namespace.c segfaulting due to deref nullptr, and I fix this here.  I don't understand code base well enough for proper fix, but I guess this should not do any harm.

### DIFF
--- a/src/lib-storage/mail-namespace.c
+++ b/src/lib-storage/mail-namespace.c
@@ -735,7 +735,7 @@ mail_namespace_find_unsubscribable(struct mail_namespace *namespaces,
 struct mail_namespace *
 mail_namespace_find_inbox(struct mail_namespace *namespaces)
 {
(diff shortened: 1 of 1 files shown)


sirainen:
Clarified the API behavior with https://github.com/dovecot/core/commit/bb1da4a6320eec11890c852e74a08868837e7be3 - So although this change fixed the one specific case it's mostly just luck that the caller allowed NULL to be returned. In earlier code it could have returned NULL, but it was becoming annoying and nowadays there's this guarantee. So the crash you saw should rather be fixed at the caller level.

PR #9, opened 2016-04-19
Add a new type of ACL to permit a fake empty mailbox

alexkiro:
The "fake-empty" ACL permission works exactly like the "read" ACL
permission. This permits selecting, searching and fetching from the
mailbox. However the mailbox will always appear empty even though
messages are in the mailbox.

This particular permission is useful when the user is granted the
permission to append messages to the mailbox, but not read any
messages. Some mail clients are not able to handle such configurations.
(for example outlook 2013).


cmouse:
Can you please clean the pull request? It contains excess commits.

sirainen:
I wonder if it's better to have this as a separate ACL as in this patch, or if this should just be a setting that applies to the regular "r" ACL. Would anyone want to use both regular "r" and "fake-empty" at the same time?

sirainen:
Also it looks like this fake-empty doesn't prevent STATUS or SELECT command from exposing the number of emails, unseen emails, etc. So this is a bit dangerous as it is now because it leaks information. So I think it would at least need to hook into get_status() and provide dummy 0 or 1 values for everything. Not sure if that is enough though. Also that needs to be tested with mailbox_list_index=yes.

alexkiro:
Thanks for the feedback.

> I wonder if it's better to have this as a separate ACL as in this patch, or if this should just be a setting that applies to the regular "r" ACL. Would anyone want to use both regular "r" and "fake-empty" at the same time?

For our use case we have users that have Read access to a mailbox, and write-only access to another. And they need to be able to move messages from the read one to the write-only one.

I think it might be better to have them mutually exclusive though. 

> Also it looks like this fake-empty doesn't prevent STATUS or SELECT command from exposing the number of emails, unseen emails, etc. So this is a bit dangerous as it is now because it leaks information. So I think it would at least need to hook into get_status() and provide dummy 0 or 1 values for everything. Not sure if that is enough though. Also that needs to be tested with mailbox_list_index=yes.

You're absolutely right, will update the pull request.

sirainen:
I'll close this for now. As mentioned, the current functionality of the patch is too dangerous outside a specific use case.

PR #10, opened 2016-05-20
dsync: race condition where done/finish is received after one side closes

bdraco:
We do not tell the remote we are closing if they have
already told us because they close the
connection after sending ITEM_FINISH or ITEM_DONE and will
not be ever receive anything else from us unless
it just happens to get combined into the same packet
as a previous response and is already in the buffer.

This solved the problem I reported here

http://www.dovecot.org/list/dovecot/2016-May/104337.html

I was getting a EPIPE about 1:30 dsync runs over the socket. I have received 0 in 10000 runs with this change.


sirainen:
Merged.

PR #12, opened 2016-07-05
Fix data stack memory leak in FTS storage layer

bechtoldt:
As `fts_transaction_commit' is a top-level plug-in hook memory allocated
on the data stack must be freed.


sirainen:
I don't think this is needed? Data stack is intended to be used somewhat lazily so it can leak for a little bit. It's freed before returning from mailbox_transaction_commit().

bechtoldt:

PR #13, opened 2016-07-05
Fix segfault when preparing BODYSTRUCTURE response

bechtoldt:
Apparently the `context' member of a `message_part_body_data' item may become NULL supposedly when fetching BODYSTRUCTURE for mails with broken MIME structure. None of the code here really is prepared for this situation and so this might apply to FETCH BODY responses as well.


sirainen:
Hmm. I'm not sure if this is the right fix for it. More likely I'd want here to be i_assert(part->context != NULL) and fix others parts of code to make sure it never is. Is there some test mail that can be used to reproduce the crash?

bechtoldt:

PR #14, opened 2016-07-05
Properly merge header fields

bechtoldt:
In `index_mail_add_temp_wanted_fields' properly merge header fields by dropping duplicates rather than appending them over and over again.
Time complexity could be improved as this now performs in O(n^2) for normally small n, though.


sirainen:
I was assuming that mailbox_header_lookup_init() did the deduplication, but looking at its code I don't think it does. So this deduplication could be moved there and it could help with other parts of the code as well. Also it already sorts the headers so it would also avoid the O(n^2).

bechtoldt:

PR #15, opened 2016-07-05
Add new config options for fts_solr

bechtoldt:
"From", "To", "Cc", "Bcc", "Subject"

The default for sending header fields to Solr is to send a limited number of headers ("From", "To", "Cc", "Bcc", "Subject") as "field" attributes and all other headers with name and content in the "hdr" field.

```
<field name="From">from@example.com</field>
<field name="To">to@example.com</field>
..
<field name="hdr">
From: from@example.com
To: to@example.com
...
X-Origin-Reference: foo1
X-Origin: foo2
X-Origin-Recipient: foo3
X-Priority: 3
...
X-Foo: content of x-foo
</field>
```

The new config switches to the following:
limit_mime_hdr: limit fields listed in "hdr" field to the same fields as in the standard fields header (From, To, Cc, Bcc, Subject):

```
<field name="From">from@example.com</field>
<field name="To">to@example.com</field>
..
<field name="hdr">
From: from@example.com
To: to@example.com
</field>
skip_mime_hdr_fieldname: Do not put the name of the field in the "hdr" field
<field name="From">from@example.com</field>
<field name="To">to@example.com</field>
..
<field name="hdr">
from@example.com
to@example.com
...
foo1
foo2
foo3
3
...
content of x-foo
</field>
```

if you set both of the config values the result will be something like this:

```
<field name="From">from@example.com</field>
<field name="To">to@example.com</field>
..
<field name="hdr">
from@example.com
to@example.com
...
</field>
```

To use this config values add them to the fts_solr parameter list:

```
plugin {
  # Search
  fts = solr
  fts_solr = break-imap-search url=http://solr.example.com:8080/solr/collection1/ limit_mime_hdr skip_mime_hdr_fieldname
}
```

### DIFF
--- a/src/plugins/fts-solr/fts-backend-solr.c
+++ b/src/plugins/fts-solr/fts-backend-solr.c
@@ -510,21 +510,39 @@ fts_backend_solr_update_set_build_key(struct fts_backend_update_context *_ctx,
 {
 	struct solr_fts_backend_update_context *ctx =
 		(struct solr_fts_backend_update_context *)_ctx;
+	struct solr_fts_backend *backend = (struct solr_fts_backend *)ctx->ctx.backend;
+	struct fts_solr_user *fuser = FTS_SOLR_USER_CONTEXT(backend->backend.ns->user);
 
 	if (key->uid != ctx->prev_uid)
 		fts_backend_solr_uid_changed(ctx, key->uid);
 
+	bool want_indexed = FALSE;
+	bool want_hdr = TRUE;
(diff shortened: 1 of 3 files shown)


sirainen:
These changes make using fts-solr even less IMAP-compliant though.. For example I think it now breaks trying to search Message-IDs? That's at least somewhat commonly used by IMAP clients (e.g. iOS is using it excessively nowadays for some weird reason).

bechtoldt:

PR #16, opened 2016-07-15
Fix searching multiple mailboxes with fts_solr

x16a0:
When searching multiple mailboxes with fts_solr, a condition to search these mailboxes is added to the query string. However, this condition has to be separated from the preceding condition by a space (a '+' in this case, as it's encoded); otherwise, Solr considers it to be part of the previous condition, in which case it will fail to parse it properly. Therefore, this patch adds that one missing character so it works again.


cmouse:
Hi, we are doing internal review on this

sirainen:
Merged.

PR #18, opened 2016-08-20
Set MemoryDenyWriteExecute=true in systemd service

candrews:
Introduced in systemd 231, MemoryDenyWriteExecute=true causes attempts to create memory mappings that are writable and executable at the same time, or changes to existing memory mappings to become executable, to be prohibited.


candrews:
Note that on systemd < 231, this option is ignored.

cmouse:
We'll consider this.

cmouse:
This seems too scary to add by default, admins that are concerned can add this in their own drop-ins.

PR #21, opened 2016-10-10
Allow setting elliptic curves using ssl_curve_list parameter

juhakoho:
Hello,

Please consider this PR that will add the possibility to set the elliptic curves using a new parameter "ssl_curve_list".

### DIFF
--- a/src/lib-ssl-iostream/iostream-openssl.c
+++ b/src/lib-ssl-iostream/iostream-openssl.c
@@ -162,7 +162,7 @@ openssl_iostream_set(struct ssl_iostream *ssl_io,
 	}
 #if OPENSSL_VERSION_NUMBER >= 0x10002000L
 	if (set->curve_list != NULL && strlen(set->curve_list) > 0 &&
-		strcmp(ctx_set->curve_list, set->curve_list) != 0) {
+		(ctx_set->curve_list == NULL || strcmp(ctx_set->curve_list, set->curve_list) != 0)) {
 		if (SSL_set1_curves_list(ssl_io->ssl, set->curve_list) == 0) {
 			*error_r = t_strdup_printf(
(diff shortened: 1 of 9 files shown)


sirainen:
This would at least need some #ifdef to build with older OpenSSL versions. Also at the same time it should be added to lib-ssl-iostream (the ssl-proxy-openssl.c code duplication is going away in v2.3).

juhakoho:
Thanks for clearing this up. I was a bit confused about lib-ssl-iostream and duplication in ssl-proxy-openssl.

The new commit I suppose fixes the issues you mentioned.

src/lib-ssl-iostream/iostream-openssl.c
```diff
@@ -160,6 +160,17 @@ openssl_iostream_set(struct ssl_iostream *ssl_io,
 			return -1;
 		}
 	}
+#if OPENSSL_VERSION_NUMBER >= 0x10002000L
+	if (set->curve_list != NULL && strlen(set->curve_list) > 0 &&
+		strcmp(ctx_set->curve_list, set->curve_list) != 0) {
```
sirainen: It could be that set->curve_list != NULL, but ctx_set->curve_list == NULL and this would crash then.

juhakoho: Good point... fixed. At least I assume that ctx_set->curve_list being NULL should behave like if the strings are not equal.

cmouse:
hi! can you rebase your work?

juhakoho:
Hi, sorry for the late response but I was on holidays. Should be okay now.

mrannanj:
A slightly modified version of this is in internal review.

juhakoho:
Great! Thanks!

sirainen:
Merged

PR #22, opened 2016-10-28
configure: Improve check for OpenSSL without EC support

pghmcfc:
The original test was for EC_KEY_new but some systems had that and not
EVP_PKEY_CTX_new_id, so the test was switched to that function.
However, Fedora releases 12 through 17 have EVP_PKEY_CTX_new_id but
not EC_KEY_new. So we need to test for both functions before enabling
the dcrypt build.


sirainen:
merged

PR #23, opened 2016-10-31
configure: Fix build with old OpenSSL without SSL_clear_options

pghmcfc:
SSL_clear_options was introduced in OpenSSL 0.9.8m but may be
backported to older versions in "enterprise" OS releases, so a version
check is insufficient here.

It was originally implemented as a macro but is a function in more
recent OpenSSL versions, so a test that works for both cases is needed.


cmouse:
Hi, we are taking this into review

sirainen:
Merged

PR #24, opened 2016-12-01
config: Match multiple names in local_name

bdraco:
This can significantly reduce memory usage when using
a UCC certificate with multiple names by only loading
the certificate and key once.

### DIFF
--- a/src/config/config-filter.c
+++ b/src/config/config-filter.c
@@ -37,7 +37,17 @@ static bool config_filter_match_rest(const struct config_filter *mask,
 	if (mask->local_name != NULL) {
 		if (filter->local_name == NULL)
 			return FALSE;
-		if (dns_match_wildcard(filter->local_name, mask->local_name) != 0)
+		/* Handle multiple names seperated by spaces in local_name
+		 * Ex: local_name "mail.domain.tld domain.tld mx.domain.tld" { ... } */
+		const char *const *local_name = t_strsplit_spaces(mask->local_name, " ");
+		bool matched = FALSE;
+		for (; *local_name != NULL; local_name++) {
+			if (dns_match_wildcard(filter->local_name, *local_name) == 0) {
+				matched = TRUE;
+				break;
+			}
+		}
+		if (!matched)
 			return FALSE;
 	}
 	/* FIXME: it's not comparing full masks */


cmouse:
Thanks! We'll submit this for review.

src/config/config-filter.c
```diff
@@ -37,8 +37,23 @@ static bool config_filter_match_rest(const struct config_filter *mask,
 	if (mask->local_name != NULL) {
 		if (filter->local_name == NULL)
 			return FALSE;
-		if (dns_match_wildcard(filter->local_name, mask->local_name) != 0)
+		/* Handle multiple names seperated by spaces in local_name
+		 * Ex: local_name "mail.domain.tld domain.tld mx.domain.tld" { ... } */
+		if (strstr(mask->local_name, " ") != NULL) {
```
cmouse: this if is useless, t_strsplit_spaces will return valid array in any case, so you could just leave the code below and remove the if completely.

bdraco: done

bdraco:
Changed

bdraco:
Confirmed working as expected with the change

Added 
local_name "doesitblend.org secondname.tld" {

`$ openssl s_client -connect localhost:993 -showcerts -servername secondname.tld
CONNECTED(00000003)
depth=3 C = SE, O = AddTrust AB, OU = AddTrust External TTP Network, CN = AddTrust External CA Root
verify return:1
depth=2 C = GB, ST = Greater Manchester, L = Salford, O = COMODO CA Limited, CN = COMODO RSA Certification Authority
verify return:1
depth=1 C = US, ST = TX, L = Houston, O = "cPanel, Inc.", CN = "cPanel, Inc. Certification Authority"
verify return:1
depth=0 CN = doesitblend.org`

`$ openssl s_client -connect localhost:993 -showcerts -servername doesitblend.org
CONNECTED(00000003)
depth=3 C = SE, O = AddTrust AB, OU = AddTrust External TTP Network, CN = AddTrust External CA Root
verify return:1
depth=2 C = GB, ST = Greater Manchester, L = Salford, O = COMODO CA Limited, CN = COMODO RSA Certification Authority
verify return:1
depth=1 C = US, ST = TX, L = Houston, O = "cPanel, Inc.", CN = "cPanel, Inc. Certification Authority"
verify return:1
depth=0 CN = doesitblend.org`

src/config/config-filter.c
```diff
@@ -37,7 +37,17 @@ static bool config_filter_match_rest(const struct config_filter *mask,
 	if (mask->local_name != NULL) {
 		if (filter->local_name == NULL)
 			return FALSE;
-		if (dns_match_wildcard(filter->local_name, mask->local_name) != 0)
+		/* Handle multiple names seperated by spaces in local_name
+		 * Ex: local_name "mail.domain.tld domain.tld mx.domain.tld" { ... } */
+		const char *const *local_name = t_strsplit_spaces(mask->local_name, " ");
+		int matched = 0;
```
cmouse: this needs to be bool

cmouse: this needs to be bool, not int.

src/config/config-filter.c
```diff
@@ -37,7 +37,17 @@ static bool config_filter_match_rest(const struct config_filter *mask,
...
+		for (; *local_name != NULL; local_name++) {
+			if (dns_match_wildcard(filter->local_name, *local_name) == 0) {
+				matched = 1;
```
cmouse: use TRUE/FALSE.

src/config/config-filter.c
bdraco: Yes it definitely should be, will push as soon as the new rpm finishes + retest

bdraco:
Confirmed working as expected with the bool change

Added 
local_name "doesitblend.org secondname.tld" {

`$ openssl s_client -connect localhost:993 -showcerts -servername secondname.tld
CONNECTED(00000003)
depth=3 C = SE, O = AddTrust AB, OU = AddTrust External TTP Network, CN = AddTrust External CA Root
verify return:1
depth=2 C = GB, ST = Greater Manchester, L = Salford, O = COMODO CA Limited, CN = COMODO RSA Certification Authority
verify return:1
depth=1 C = US, ST = TX, L = Houston, O = "cPanel, Inc.", CN = "cPanel, Inc. Certification Authority"
verify return:1
depth=0 CN = doesitblend.org`

`$ openssl s_client -connect localhost:993 -showcerts -servername doesitblend.org
CONNECTED(00000003)
depth=3 C = SE, O = AddTrust AB, OU = AddTrust External TTP Network, CN = AddTrust External CA Root
verify return:1
depth=2 C = GB, ST = Greater Manchester, L = Salford, O = COMODO CA Limited, CN = COMODO RSA Certification Authority
verify return:1
depth=1 C = US, ST = TX, L = Houston, O = "cPanel, Inc.", CN = "cPanel, Inc. Certification Authority"
verify return:1
depth=0 CN = doesitblend.org`


Also confirmed no regression on a single name: mail.myssltest.org

`openssl s_client -connect localhost:993 -showcerts -servername mail.myssltest.org
CONNECTED(00000003)
depth=3 C = SE, O = AddTrust AB, OU = AddTrust External TTP Network, CN = AddTrust External CA Root
verify return:1
depth=2 C = GB, ST = Greater Manchester, L = Salford, O = COMODO CA Limited, CN = COMODO RSA Certification Authority
verify return:1
depth=1 C = US, ST = TX, L = Houston, O = "cPanel, Inc.", CN = "cPanel, Inc. Certification Authority"
verify return:1
depth=0 OU = Domain Control Validated, OU = PositiveSSL Wildcard, CN = *.myssltest.org
verify return:1`

bdraco:
Thanks for the quick reviews.

cmouse:
merged.

chris001:
Was this feature added to 2.2.27 ?

cmouse:
https://github.com/dovecot/core/commits/release-2.2.27 , yep.

chris001:
Looking at adding code to virtualmin to generate config file lines by extracting hostnames from the cert or certs associated with virtual servers enabled for Dovecot mail reading.  It occurs to me that the domain names could be, and probably should be, programmatically extracted automatically from the TLS cert, directly by Dovecot, with a couple of lines of code.  Would it not make more sense, instead of a user or a control panel having to explicitly spell out, or extract host names from a cert,  to tell Dovecot, "I am specifying this TLS cert, so from now on, all hostnames contained in this cert are eligible for recognition during the STARTTLS SNI phase." ??

cmouse:
It's not actually trivial to extract the names in config phase, since it requires something that actually understands X.509 data structure. The current way things are done this is not really "a few lines of code" change.

cmouse:
This is not saying we do not intend to fix this in some other way in the future.

chris001:
OK.  It looks like `openssl` extracts any x509 info.  This could probably be used.
```
:~# openssl x509 -noout -subject -in /etc/letsencrypt/live/mydomain.com/cert.pem
subject= /CN=mydomain.com
```

cmouse:
or probably not.

chris001:
Possibly for both virtualmin and dovecot?  It seems the right thing to do, to tell Dovecot, "when the client tries to do TLS SNI, use these certs or these folders of certs, and be smart, match whichever cert contains the domain name which the client tries to Server Name Identify".  Instead of forcing the control panel or user to spell it all out, and be prone to mis-specifying it, making whichever errors.

cmouse:
Yes, it's not possible today to be "smart" in dovecot about this. It's not a matter of slabbing in system("openssl x509 ...") and parsing the output. I understand your pain about this, but today it's not doable. Sorry.

chris001:
OK.  Well, is there any particular reason why you say it's not doable today, to add this very simple feature, for dovecot to obtain SNI TLS certs from a folder tree that would be specified in `dovecot.conf`, for example, `/etc/letsencrypt/live`?

cmouse:
we are happy to accept a pull request! =)

PR #25, opened 2016-12-05
configure: Fix HAVE_MYSQL_SSL_CIPHER check

pghmcfc:
Detection of mysql_ssl_set() with a cipher parameter is broken due to a typo.


cmouse:
Thanks, we'll look into it.

cmouse:
merged.

PR #26, opened 2016-12-05
Skip mail key tests if dcrypt not available

pghmcfc:
Same methodology as in the lib-dcrypt tests.

These tests currently crash in the absence of ECC support in OpenSSL.


cmouse:
Thanks, we'll look into this.

cmouse:
Fixed slightly differently, but thanks!

PR #27, opened 2016-12-05
configure: Fix some implicit function declarations

pghmcfc:
Some configure tests fail unexpectedly if the compiler flag
-Werror=implicit-function-declarations is enabled, which can result
in the wrong implementations being used.

This compiler flag is now enabled by default in Fedora Rawhide:
https://fedoraproject.org/wiki/Changes/Fedora26CFlags

<stdlib.h> is needed for exit()
<string.h> is needed for strcpy()


cmouse:
Hi! We'll take a look at this.

cmouse:
merged.

PR #28, opened 2016-12-06
config: Reduce memory required to load config filters

bdraco:
Create a new data_stack_frame in config_filter_find_all
in order to ensure the memory allocated in
config_filter_match_rest is freed

Optimize handling local_names by avoiding calls to
t_strsplit_spaces (which allocates memory) when there
are no spaces in the local_name

Testing:

Before changes.  4825 local_names took 6.13 GiB RSS
After changes.   4825 local_names took 536 MiB RSS

### DIFF
--- a/src/config/config-filter.c
+++ b/src/config/config-filter.c
@@ -204,11 +204,18 @@ config_filter_find_all(struct config_filter_context *ctx, pool_t pool,
 	ARRAY_TYPE(config_filter_parsers) matches;
 	ARRAY_TYPE(const_string) service_names;
 	unsigned int i;
+	unsigned int t_id;
 
 	memset(output_r, 0, sizeof(*output_r));
 
 	p_array_init(&matches, pool, 8);
 	p_array_init(&service_names, pool, 8);
+
+	/* create a new stack after matches has been allocated
+	 * so we can free the memory used by config_filter_match_rest
+	 * with the below t_pop call */
+	t_id = t_push("config_filter_find_all");
+
 	for (i = 0; ctx->parsers[i] != NULL; i++) {
 		const struct config_filter *mask = &ctx->parsers[i]->filter;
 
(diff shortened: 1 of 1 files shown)


cmouse:
Thanks, we'll take a look at this. Impressive numbers.

src/config/config-filter.c
```diff
@@ -204,11 +212,18 @@ config_filter_find_all(struct config_filter_context *ctx, pool_t pool,
...
 
 	memset(output_r, 0, sizeof(*output_r));
 
 	p_array_init(&matches, pool, 8);
 	p_array_init(&service_names, pool, 8);
+
+	/* create a new stack after matches has been allocated
+	 * so we can free the memory used by config_filter_match_rest
+	 * with the below t_pop call */
+	t_id = t_push_named("config_filter_find_all");
```
cmouse: you can use T_PUSH / T_POP instead.

src/config/config-filter.c
```diff
@@ -38,17 +38,25 @@ static bool config_filter_match_rest(const struct config_filter *mask,
...
-		bool matched = FALSE;
-		for (; *local_name != NULL; local_name++) {
-			if (dns_match_wildcard(filter->local_name, *local_name) == 0) {
-				matched = TRUE;
-				break;
+		 * Ex: local_name "mail.domain.tld domain.tld mx.domain.tld" { ... }
+		 * We check to see if the string has a " " in it so we can avoid
+		 * memory pool allocation in t_strsplit_spaces
+		 */
+		if (strstr(mask->local_name, " ") != NULL) {
```
cmouse: please do not replace this code, the previous code was better.

bdraco: I'm happy to take it out, however it really is making a noticeable difference

`Test #1 compiled without ststr
# time doveconf 

real	0m11.877s
user	0m11.528s
sys	0m0.321s

Time #2 compiled with ststr
# time doveconf 
real	0m6.438s
user	0m6.090s
sys	0m0.329s

# grep local_name /etc/dovecot/sni.conf | wc
   4825   14475  161207`

src/config/config-filter.c
cmouse: Well, I suppose this goes as well, but I ment literally T_PUSH { config_filter_find_all() } T_POP;

cmouse:
Fixed differently. Thanks for the suggestion.

bdraco:
@cmouse  Thank you.

PR #29, opened 2016-12-07
openssl: Clear error queue after an incomplete SSL_shutdown

manuelm:
If the SSL_shutdown-call fails (e.g. because the underlaying socket has
already been closed) OpenSSL puts the corresponding error into the
queue. We don't care about details so we need to clear the queue.

Otherwise the error will be pulled while error checking the next OpenSSL
call of an unrelated connection.

Issue can be reproduced by:
* setting `service imap-login { service_count = 0 }`
* imap-connection 1 (TLS) is doing a NOOP every second
* tcp-connection 2 is connecting to and immediately disconnecting from the SSL/TLS port (without any handshake).

### DIFF
--- a/src/lib-ssl-iostream/iostream-openssl.c
+++ b/src/lib-ssl-iostream/iostream-openssl.c
@@ -299,7 +299,11 @@ static void openssl_iostream_unref(struct ssl_iostream *ssl_io)
 
 static void openssl_iostream_destroy(struct ssl_iostream *ssl_io)
 {
-	(void)SSL_shutdown(ssl_io->ssl);
+	if (SSL_shutdown(ssl_io->ssl) != 1) {
+		/* if bidirectional shutdown fails we need to clear
+		   the error queue */
+		openssl_iostream_clear_errors();
+	}
 	(void)openssl_iostream_more(ssl_io);
 	(void)o_stream_flush(ssl_io->plain_output);
 	/* close the plain i/o streams, because their fd may be closed soon,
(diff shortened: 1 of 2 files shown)


cmouse:
Thanks we'll take a look at this.

BinaryKhaos:
A related change in OpenSSL (SSL_shutdown() no longer returns success if called early in the handshake, i.e. connection closed early and the handshake never went through) that triggered problems in Dovecot:
https://github.com/openssl/openssl/commit/64f9f40696f993406e53c16d7c9d815004afd8ad

sirainen:
I'm a bit worried about this: "because the underlaying socket has already been closed" - do you see EBADF errors because OpenSSL is trying to access an already-closed socket fd? If yes, that is still a bug that should be fixed, since with bad luck maybe the socket is already used by something else and that unrelated socket is disconnected.

manuelm:
As far as I remember in my example above it's `ECONNRESET`. OpenSSL then puts `SSL_R_SHUTDOWN_WHILE_IN_INIT` into the error queue. And the error checking in the next `SSL_read`-loop will revive the unrelated error from the queue and bail out.

So with closed I'm mean the other side has closed the connection and the bidirectional SSL shutdown can't happen any more.

manuelm:
I've changed the title to better reflect the actual change. Dropping the unrelated connection is the symptom. If desired I can amend the commit as well.

cmouse:
Please do amend, it's better if the commit message correctly describes the change.

manuelm:
amended

cmouse:
merged.

PR #32, opened 2017-01-04
auth-policy: hashed_password will always be blank, tell buffer it has…

bearggg:
… data so str_len works when converting to hex

### DIFF
--- a/src/auth/auth-policy.c
+++ b/src/auth/auth-policy.c
@@ -453,7 +453,7 @@ void auth_policy_create_json(struct policy_lookup_ctx *context,
 		digest->loop(ctx, password, strlen(password));
 	ptr = (unsigned char*)str_c_modifiable(buffer);
 	digest->result(ctx, ptr);
-	str_truncate(buffer, digest->digest_size);
+	buffer_set_used_size(buffer, digest->digest_size);
 	if (context->set->policy_hash_truncate > 0) {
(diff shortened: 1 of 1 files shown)


cmouse:
Um. Those two are the same thing.

bearggg:
It does look like it, however the str->used is zero at that point, so when str_truncate runs the if statement doesnt actually run buffer_set_used_size.

```
/* Truncate the string to specified length. If it's already smaller,
   do nothing. */
static inline void str_truncate(string_t *str, size_t len)
{
    if (str_len(str) > len)
        buffer_set_used_size(str, len);
}
```
Possibly the issue is further up the stack, but I dont see anything here that sets str->used? I didnt look to closely at digest->result as its just a pointer going in.
```
    ptr = (unsigned char*)str_c_modifiable(buffer);
    digest->result(ctx, ptr);
```
either way, str_len(buffer) is always zero for me, so the password word hash is always blank.  The proposed change seems to fix it and still appears safe.

sirainen:
Oh, this was somewhat recently broken by e0e555a09a8ff93aa904586aeffd37f5b0e3a84d. But yeah, the buffer there isn't really a string so it would be even cleaner to have it be buffer_t instead of string_t.

cmouse:
Opened internally, thanks.

PR #33, opened 2017-01-12
Execute /etc/sysconfig/dovecot on init, if it exists

tomsommer:
Helps users set `ulimit`, `DAEMON` etc. without having to touch the init.d script.

### DIFF
--- a/doc/dovecot-initd.sh
+++ b/doc/dovecot-initd.sh
@@ -19,6 +19,8 @@ DAEMON=/usr/local/sbin/dovecot
 # Uncomment to allow Dovecot daemons to produce core dumps.
 #ulimit -c unlimited
 
+[ -e /etc/sysconfig/dovecot ] && . /etc/sysconfig/dovecot
+
 test -x $DAEMON || exit 1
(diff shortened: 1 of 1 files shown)


cmouse:
This is pretty redhat-specific change, in debian the correct folder would be /etc/default/dovecot

tomsommer:
You want a debian entry as well?

donmichelangelo:
Isn't this something that should be left to the official package maintainers of the corresponding distributions? I mean this is pretty much contradicting to the philosophy of such binary distributions like CentOS (actually also Debian and others) if you install Dovecot straight from the source.

Ok, it's not forbidden, but it doesn't make much sense to be honest, especially if you can use the Fedora src rpm balls to compile newer Dovecot versions for RHEL 6/7 targets.

tomsommer:
Well, some of us compile from source and copy the bundled init script in - that's why it's there?

cmouse:
In internal review

sirainen:
Could have been both in the same commit, but merged anyway.

PR #34, opened 2017-01-12
[WIP] dovecot-rda

manuelm:
For my current work I need to consolidate the mail storage backend for a couple of servers to one server while keeping the MTA in place. The MTA is currently using `dovecot-lda` and doesn't support LMTP so I came up with a drop-in replacement I simply called `dovecot-rda`.

I could have used pythons `smtplib` or some other LMTP library/client but this would require me to at least add support for rejection mails and the logging format.

I'm not sure this is something you want to merge, so feel free to say No here.

In case you think this is a nice addition to `dovecot-lda` please don't hesitate to leave comments. I want to get this into a merge-able state.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -892,6 +892,7 @@ src/config/Makefile
 src/doveadm/Makefile
 src/doveadm/dsync/Makefile
 src/lda/Makefile
+src/rda/Makefile
 src/log/Makefile
 src/lmtp/Makefile
 src/dict/Makefile
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -53,6 +53,7 @@ SUBDIRS = \
 	pop3-login \
 	pop3 \
 	lda \
+	rda \
 	lmtp \
 	log \
 	config \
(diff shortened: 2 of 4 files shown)


sirainen:
I've thought about adding support for LMTP client-side delivery to Dovecot, but haven't been sure what would be the nicest way to do it. Maybe a parameter to dovecot-lda or maybe some new doveadm command. I guess it could also be a new write-only mail storage backend, which might make it more flexible, at least in some ways, but probably worse in others.. In any case your current patch has a bit too much copy&pasting :)

manuelm:
> Maybe a parameter to dovecot-lda or maybe some new doveadm command

I first looked into adding this as a parameter to `dovecot-lda` but quickly thought it doesn't fit that well. Apart from the wrong name of the tool, some parameters don't make sense in LMTP/SMTP delivery context. That said I don't have much of a problem to add a `-R protocol://host:port` (or something else) to `dovecot-lda` if that's preferred.

> In any case your current patch has a bit too much copy&pasting :)

Yeah I know that it includes a couple of functions from lda because that's were I started (and ended). Code sharing and adding a man page is on my todo list. But before spending more work on this I first wanted to hear if it is of any interest.

PR #35, opened 2017-01-26
fts-solr: Allow username/password for solr connction

bdraco:
Solr comes open to the world by default. For better
security, we protect our solr installs with a user
and password.  This change allows dovecot to connect
to these instances.

Example usage in a fts.conf (permissions 0600)
 fts_solr = url=http://nick:pass@127.0.0.1:8983/solr/dovecot/

### DIFF
--- a/src/plugins/fts-solr/solr-connection.c
+++ b/src/plugins/fts-solr/solr-connection.c
@@ -62,6 +62,8 @@ struct solr_connection {
 	in_port_t http_port;
 	char *http_base_url;
 	char *http_failure;
+	char *http_user;
+	char *http_password;
 
 	int request_status;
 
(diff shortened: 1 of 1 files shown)


src/plugins/fts-solr/solr-connection.c
```diff
@@ -151,6 +159,10 @@ void solr_connection_deinit(struct solr_connection **_conn)
 	XML_ParserFree(conn->xml_parser);
 	i_free(conn->http_host);
 	i_free(conn->http_base_url);
+	if (conn->http_user != NULL) {
+		i_free(conn->http_user);
+		i_free(conn->http_password);
+	}
```
mrannanj: i_free is NULL safe I believe.

cmouse:
Hi! We'll take this into consideration.

cmouse:
Merged with small changes.

bdraco:
Thanks

PR #37, opened 2017-02-26
Spelling

jsoref:
I'd imagine you might want this split. Just provide instructions.

### DIFF
--- a/src/lib-dcrypt/ostream-encrypt.c
+++ b/src/lib-dcrypt/ostream-encrypt.c
@@ -1,4 +1,4 @@
-/* file truct dcrypt_public_keyyntax
+/* file truct dcrypt_public_key syntax
  * magic (14 bytes)
  * version (1 bytes)
  * flags (4 bytes)
--- a/src/lib-dcrypt/ostream-encrypt.c
+++ b/src/lib-dcrypt/ostream-encrypt.c
@@ -1,4 +1,4 @@
-/* file truct dcrypt_public_key syntax
+/* file struct dcrypt_public_key syntax
  * magic (14 bytes)
  * version (1 bytes)
  * flags (4 bytes)
--- a/src/lib-storage/mailbox-attribute.c
+++ b/src/lib-storage/mailbox-attribute.c
@@ -380,7 +380,7 @@ mailbox_attribute_iter_init(struct mailbox *box,
 
 	/* copy relevant attributes */
 	array_foreach(&extra_attrs, attr) {
-		/* skip internal server attributes unless we're interating inbox */
+		/* skip internal server attributes unless we're iterating inbox */
 		if (!box->inbox_any &&
 		    strncmp(*attr, MAILBOX_ATTRIBUTE_PREFIX_DOVECOT_PVT_SERVER,
 			    strlen(MAILBOX_ATTRIBUTE_PREFIX_DOVECOT_PVT_SERVER)) == 0)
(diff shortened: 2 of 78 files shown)


cmouse:
Not necessarely. These are rather minute changes  and probably can be taken as is. But we'll look into this.

cmouse:
Or maybe following changes:
- separate commit for docs
- change the message to 'global: Fix spelling' for the code commit changes

src/lib-fts/fts-language.c
```diff
@@ -13,8 +13,8 @@
 #endif
 
 #ifndef TEXTCAT_RESULT_UNKNOWN /* old textcat.h has typos */
-#  ifdef TEXTCAT_RESULT_UNKOWN
-#    define TEXTCAT_RESULT_UNKNOWN TEXTCAT_RESULT_UNKOWN
+#  ifdef TEXTCAT_RESULT_UNKNOWN
+#    define TEXTCAT_RESULT_UNKNOWN TEXTCAT_RESULT_UNKNOWN
```
sirainen: These are intentional typos and must not be changed.

src/lib-settings/settings-parser.h
```diff
@@ -179,9 +179,9 @@ bool settings_check(const struct setting_parser_info *info, pool_t pool,
 void settings_parse_set_expanded(struct setting_parser_context *ctx,
 				 bool is_expanded);
 /* Mark all the parsed settings with given keys as being already expanded. */
-void settings_parse_set_key_expandeded(struct setting_parser_context *ctx,
+void settings_parse_set_key_expanded(struct setting_parser_context *ctx,
 				       pool_t pool, const char *key);
-void settings_parse_set_keys_expandeded(struct setting_parser_context *ctx,
+void settings_parse_set_keys_expanded(struct setting_parser_context *ctx,
 					pool_t pool, const char *const *keys);
```
sirainen: This changes the API. For v2.2 need to verify that these aren't used outside Dovecot core. A quick grep didn't show anything. In any case this should be in its own commit.

jsoref:
@sirainen: dropped + split
@cmouse: if you don't need me to split between comments and code, that's ideal since I can focus on other things (e.g. openldap). otoh, if you do need that split, it's certainly manageable. Just let me know (preferably after you've identified anything else that needs to be dealt w/, as it's much harder to drop things once they're squashed).

And thanks for the prompt responses.

src/lib-dcrypt/ostream-encrypt.c
```diff
@@ -1,4 +1,4 @@
-/* file truct dcrypt_public_keyyntax
+/* file struct dcrypt_public_key syntax
```
jsoref: I'm really not certain about this correction. The corrupt content was present when it was introduced.
@cmouse: Did I guess correctly?

jsoref:
https://github.com/dovecot/core/blob/a825281/src/lib-storage/index/index-storage.c#L942

> 		/* the only reason we're doing the destination lookup is to
> 		   make sure that the cache file is opened and the cache
> 		   decisinos are up to date */

I could not figure out what `decisinos` was. If someone can suggest a correction, I can roll it into this PR.

sirainen:
decisions

src/lib-dcrypt/ostream-encrypt.c
cmouse: yes

cmouse:
working on this, bear with us.

src/lib-storage/mailbox-attribute.c
```diff
@@ -380,7 +380,7 @@ mailbox_attribute_iter_init(struct mailbox *box,
 
 	/* copy relevant attributes */
 	array_foreach(&extra_attrs, attr) {
-		/* skip internal server attributes unless we're interating inbox */
+		/* skip internal server attributes unless we're interacting inbox */
```
jeffpc: This should be "iterating".

cmouse:
this was refactored and merged.

PR #39, opened 2017-03-02
Export per-service connection counts, and allow more configurable prefix

via:
This implements two changes:
* Remove the hardcoded dovecot. prefix for carbon, as some organizations need specific carbon prefixes
* Split the global connection count into a hash table that has per-service connection counts, allowing carbon metrics for individual imap, pop3, lmtp, sieve, doveadm, etc... connection counts.

### DIFF
--- a/src/stats/client-export.c
+++ b/src/stats/client-export.c
@@ -499,7 +499,7 @@ static int client_export_iter_global(struct client *client)
 	str_printfa(cmd->str, "%ld", (long)g->reset_timestamp);
 	client_export_timeval(cmd->str, &g->last_update);
 	str_printfa(cmd->str, "\t%u\t%u\t%u\t",
-		    g->num_logins, g->num_cmds, g->num_connected_sessions);
+		    g->num_logins, g->num_cmds, g->num_connected_sessions_combined);
 	client_export_stats(cmd->str, g->stats);
 	str_append_c(cmd->str, '\n');
 	o_stream_nsend(client->output, str_data(cmd->str),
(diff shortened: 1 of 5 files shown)


cmouse:
Oh wow, thanks! We'll look into this definetly.

cmouse:
Could you split the commit into two, so that there will be one commit about the prefix configuration, and another about the per-service connection counts?

There were also some trailing whitespace.

via:
The prefix commit is now split off, trailing whitespace removed.

Are there thoughts on using the default pool to allocate the uint for the count and strdup for the service names in the hash table? I don't bother to free them since they should be there for the duration of the run, but I don't know if thats acceptable.

cmouse:
We usually frown upon leaving memory allocated on exit. It would be best to use i_free to release the memory on deinit.

sirainen:
One thing I've wanted for a while now is to provide full per-service stats outputs. So stats plugin would keep tracking all the stats for each service separately. When dumping global stats it would sum up these for the output (so not needing to have an explicit global stats internally). This patch is doing a little bit of the same, but only for one field. It would also make it more difficult to implement such per-service stats in a backwards compatible way..

via:
Thats fair, I guess you're thinking of another zoom level between global and domain/ip?  I could take a stab at that in my free time, though my immediate goal is to have information on how close we are to process limits submitted to graphite, and in the short term might use this patch anyway if you think it looks safe overall.

Would you consider taking the commit to remove the hardcoded prefix?

cmouse:
This is going to be done with new stats somehow.

PR #40, opened 2017-04-04
imapc: Fix memory leak in disconnected state

zeil:


src/lib-imap-client/imapc-connection.c
```diff
@@ -382,6 +382,10 @@ static void imapc_connection_set_state(struct imapc_connection *conn,
 
 		conn->selecting_box = NULL;
 		conn->selected_box = NULL;
+
+		if (conn->ips != NULL)
```
jeffpc: IIRC, `i_free(NULL)` is a no-op so there is no need for this NULL check.

jeffpc: Hrm, looking at the implementation of it, I may be wrong - it might try to memset the memory first.

sirainen: i_free(NULL) is no-op, just like free(NULL) in ANSI-C.

sirainen: ..and it's used all over the code everywhere, so better to be consistent here also and not have the if-check :)

cmouse:
Merged outside github.

PR #42, opened 2017-04-26
Deprecated function

petervanderdoes:
Function DH_generate_parameters has been deprecated since OpenSSL 0.9.8

### DIFF
--- a/src/lib-ssl-iostream/iostream-openssl-params.c
+++ b/src/lib-ssl-iostream/iostream-openssl-params.c
@@ -13,10 +13,19 @@ generate_dh_parameters(int bitsize, buffer_t *output, const char **error_r)
 {
         DH *dh;
 	unsigned char *p;
-	int len, len2;
+	int len, len2, success;
 
+#if OPENSSL_VERSION_NUMBER >= 0x00908000L
+	success = DH_generate_parameters_ex(dh, bitsize, DH_GENERATOR, NULL);
+#else
+	success = 1;
 	dh = DH_generate_parameters(bitsize, DH_GENERATOR, NULL, NULL);
 	if (dh == NULL) {
+		success = 0;
+	}
+#endif
+
+	if (success == 0) {
 		*error_r = t_strdup_printf(
 			"DH_generate_parameters(bits=%d, gen=%d) failed: %s",
 			bitsize, DH_GENERATOR, openssl_iostream_error());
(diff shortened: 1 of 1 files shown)


cmouse:
Thank you, we'll take this under internal review.

src/lib-ssl-iostream/iostream-openssl-params.c
```diff
@@ -13,10 +13,19 @@ generate_dh_parameters(int bitsize, buffer_t *output, const char **error_r)
 {
         DH *dh;
 	unsigned char *p;
-	int len, len2;
+	int len, len2, success;
 
+#if OPENSSL_VERSION_NUMBER >= 0x00908000L
+	success = DH_generate_parameters_ex(dh, bitsize, DH_GENERATOR, NULL);
```
cmouse: iostream-openssl-params.c: In function 'generate_dh_parameters':
iostream-openssl-params.c:19:10: error: 'dh' is used uninitialized in this function [-Werror=uninitialized]
  success = DH_generate_parameters_ex(dh, bitsize, DH_GENERATOR, NULL);
          ^

petervanderdoes:
Should be fixed now I hope. I'm not a C programmer so please bear with me :)

cmouse:
Can you try compile it and run it yourself? It seems to have issues.

cmouse:
ssl-params: Fatal: rename(/dovecot/tmp/inst1-lib/ssl-parameters.dat.tmp, /dovecot/tmp/inst1-lib/ssl-parameters.dat) failed: No such file or directory

petervanderdoes:
After the last commit it compiles on a Debian 9 system. Before that commit it would fail but not with the error message you show. :/ Hope it  works for you now.

To compile I ran:

    ./autogen.sh
    ./configure
    make

cmouse:
Did you try running it? =)

petervanderdoes:
No I didn't :) I'll see if I can setup a test environment.

sthen:
Surely there's no need to support 5-years-dead 0.9.7? Converting outright to the _ex function avoids the need for the #if and temporary variable, and removes code that nobody's going to actually use.

sirainen:
Dovecot v2.3 removes the DH generation entirely, so this is going to be somewhat temporary change anyway.

cmouse:
2.2 is no longer in active development, so I'm closing this. Sorry, and thank you for the effort!

PR #46, opened 2017-08-23
acl list from ldap - must be possible to have multiple values, that combined in one comma separated list

DaLiV:
attached patch to dovecot-2.2.31 1.patch
also as diff to git-master branch 2.patch
hardcoded userdb_acl_groups /as currently not seeing any other fields that need similar way/
in ldap best way for realizing user groups is dynlist - so labeledURIObject - some attribute.
user have something like
labeledURI :: ldap:///ou=DovecotGroups,dc=DOMAIN,dc=TLD,dc=root?specName?sub?(&(objectclass=posixgroup)(memberuid=XXXXX))
which will returns groupnames list where belongs specified user,
but that list have multiple values of same field ... which must be combined into one CSV list to be compatible with single-line string required by dovecot .

[1.patch.txt](https://github.com/dovecot/core/files/1246778/1.patch.txt)
[2.patch.txt](https://github.com/dovecot/core/files/1246828/2.patch.txt)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -716,6 +716,8 @@ DOVECOT_WANT_TEXTCAT
 
 DOVECOT_WANT_ICU
 
+DOVECOT_WANT_APPARMOR
+
 if test $have_lucene = no; then
   not_fts="$not_fts lucene"
 fi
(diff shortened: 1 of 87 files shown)


cmouse:
Uh. You made some mistake here =)

DaLiV:
If about github long conflict list - that not offered to accept simple patch /or i'm not found/ must compare - or button inactive ... making own fork - not wanted ... idealogically not clear how to make nice looking requests/patches here ...
if you about mine English - sorry - not the mothertongue

2.patch - that made from today cloned repo .... making of new pull request for that ? - may be ... but not possible delete current so not to duplicate .

otherwise - show me where is error ... i'm open for comment ...

DaLiV:
Forked. Hope that will be ok ...
Current pull request may be deleted.

PR #47, opened 2017-08-23
reseting -> resetting

andreasschulze:
suggested by Debian lintian


cmouse:
Thank you for your patch request, we'll look into it.

cmouse:
merged

PR #48, opened 2017-08-23
acl list from ldap - must be possible to have multiple values, that 

DaLiV:
combined in one comma separated list

hardcoded userdb_acl_groups /as currently not seeing any other fields that need similar way/
in ldap best way for realizing user groups is dynlist - so labeledURIObject - some attribute.
user have something like
labeledURI :: ldap:///ou=DovecotGroups,dc=DOMAIN,dc=TLD,dc=root?specName?sub?(&(objectclass=posixgroup)(memberuid=XXXXX))
which will returns groupnames list where belongs specified user,
but that list have multiple values of same field ... which must be combined into one CSV list to be compatible with single-line string required by dovecot .

### DIFF
--- a/src/auth/passdb-ldap.c
+++ b/src/auth/passdb-ldap.c
@@ -52,13 +52,27 @@ ldap_query_save_result(struct ldap_connection *conn,
 			auth_request_set_null_field(auth_request, name);
 			continue;
 		}
-		if (values[1] != NULL) {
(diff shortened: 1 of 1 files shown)


cmouse:
I think this should aim for more generic solution, instead of hardcoded whitelist of accepted keys.

cmouse:
closing due inactivity. thank you for your effort.

PR #49, opened 2017-08-29
lib: Faster base64_encode implementation

compmaniak:
New base64_encode implementation is 3.5 times faster

### DIFF
--- a/src/lib/base64.c
+++ b/src/lib/base64.c
@@ -45,36 +45,38 @@ static const unsigned char b64dec[256] = {
 
 void base64_encode(const void *src, size_t src_size, buffer_t *dest)
 {
+	const size_t res_size = MAX_BASE64_ENCODED_SIZE(src_size);
+	unsigned char *beg = buffer_append_space_unsafe(dest, res_size);
+	unsigned char *ptr = beg;
 	const unsigned char *src_c = src;
-	unsigned char tmp[4];
 	size_t src_pos;
 
-	for (src_pos = 0; src_size - src_pos > 2; src_pos += 3) {
-		tmp[0] = b64enc[src_c[src_pos] >> 2];
-		tmp[1] = b64enc[((src_c[src_pos] & 0x03) << 4) |
+	for (src_pos = 0; src_size - src_pos > 2; src_pos += 3, ptr += 4) {
+		ptr[0] = b64enc[src_c[src_pos] >> 2];
+		ptr[1] = b64enc[((src_c[src_pos] & 0x03) << 4) |
 				(src_c[src_pos+1] >> 4)];
-		tmp[2] = b64enc[((src_c[src_pos+1] & 0x0f) << 2) |
+		ptr[2] = b64enc[((src_c[src_pos+1] & 0x0f) << 2) |
 				((src_c[src_pos+2] & 0xc0) >> 6)];
-		tmp[3] = b64enc[src_c[src_pos+2] & 0x3f];
-		buffer_append(dest, tmp, 4);
+		ptr[3] = b64enc[src_c[src_pos+2] & 0x3f];
 	}
+
+	i_assert(ptr <= beg + res_size);
+
 	switch (src_size - src_pos) {
 	case 0:
 		break;
 	case 1:
-		tmp[0] = b64enc[src_c[src_pos] >> 2];
-		tmp[1] = b64enc[(src_c[src_pos] & 0x03) << 4];
-		tmp[2] = '=';
-		tmp[3] = '=';
-		buffer_append(dest, tmp, 4);
+		ptr[0] = b64enc[src_c[src_pos] >> 2];
+		ptr[1] = b64enc[(src_c[src_pos] & 0x03) << 4];
+		ptr[2] = '=';
+		ptr[3] = '=';
 		break;
 	case 2:
-		tmp[0] = b64enc[src_c[src_pos] >> 2];
-		tmp[1] = b64enc[((src_c[src_pos] & 0x03) << 4) |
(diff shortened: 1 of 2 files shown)


jeffpc:
Thanks for the contribution.

How did you do the performance measurements?

cmouse:
Hi!

After doing some initial benchmark tests I the speed increase is only about 1.4%?

src/lib/base64.c
```diff
@@ -46,35 +46,36 @@ static const unsigned char b64dec[256] = {
...
-			tmp[1] = b64enc[((src_c[src_pos] & 0x03) << 4) |
-					(src_c[src_pos+1] >> 4)];
-			tmp[2] = b64enc[((src_c[src_pos+1] & 0x0f) << 2) |
-					((src_c[src_pos+2] & 0xc0) >> 6)];
-			tmp[3] = b64enc[src_c[src_pos+2] & 0x3f];
-			src_pos += 3;
-			break;
-		}
-		buffer_append(dest, tmp, 4);
+	const size_t res_size = (src_size / 3 + (src_size % 3 > 0)) * 4;
```
jeffpc: There is a macro for this calculation - `MAX_BASE64_ENCODED_SIZE`.  (Yes, I noticed that it sometimes overestimates by 4 bytes.  I made note of it to fix that up.)

src/lib/base64.c
```diff
@@ -46,35 +46,36 @@ static const unsigned char b64dec[256] = {
...
+	unsigned char *ptr = buffer_append_space_unsafe(dest, res_size);
+	
+	size_t src_pos = 0;
+	for (; src_size - src_pos > 2; src_pos += 3) {
+		*ptr++ = b64enc[src_c[src_pos] >> 2];
```
jeffpc: I think it'd be worth it to avoid pointer arithmetic and use an index into the output to make it more obvious that there isn't a chance of an overflow.  I'm thinking something like:

```C
size_t src_pos = 0;
size_t dst_pos = 0;
for (; src_size - src_pos > 2; src_pos += 3, dst_pos += 4) {
        ptr[dst_pos + 0] = ...
        ptr[dst_pos + 1] = ...
        ptr[dst_pos + 2] = ...
        ptr[dst_pos + 3] = ...
}
```

And just to be on the safe side, after the loop I'd throw in an `i_assert` to make sure that `dst_pos` didn't go beyond `res_size`.  (I think `i_assert(dst_pos <= res_size)` should do.)

sirainen:
Overall, there are two changes here, so it would be nicer if there were also two commits that did the change instead of just replacing all of the code (which also has completely different coding style that unnecessarily changes the code). So 1) Move the last base64 block writing outside the for-loop, 2) use buffer_get_space_unsafe() in the beginning.

src/lib/base64.c
compmaniak: Yes, I know about the macro. Should I fix it in this pull request?

src/lib/base64.c
compmaniak: Sure, indexed access shouldn't impact optimized machine code. I'll rewrite it.

src/lib/base64.c
jeffpc: Up to you if you want to fix the macro.  If you do, it should be a separate commit.

compmaniak:
@jeffpc I wrote a test program with a loop with millions of iterations encoding a constant byte sequence to base64. It looks like this
```
buffer_t *buf = buffer_create_dynamic(default_pool, 16 * 1024);
clock_t t = clock();
for (size_t i = 0; i < 10000000; ++i) {
    buffer_set_used_size(buf, 0);
    base64_encode(sample_data, sample_size, buf);
}
t = clock() - t; // here we time delta
```

jeffpc:
Interesting! I thought the speedup would have been because of the lack of `realloc()` calls as the output buffer gets preallocated with your changes, but `buffer_set_used_size()` doesn't actually free the underlying buffer so the speedup has to come from the reduced conditional logic in the loop.

jeffpc:
The commits have an author name "compmaniak" and an author email @users.noreply.github.com. Is that correct?

src/lib/base64.c
```diff
@@ -45,36 +45,41 @@ static const unsigned char b64dec[256] = {
 
 void base64_encode(const void *src, size_t src_size, buffer_t *dest)
 {
+	const size_t res_size = MAX_BASE64_ENCODED_SIZE(src_size);
+	unsigned char *beg = buffer_append_space_unsafe(dest, res_size);
```
jeffpc: "beg" isn't much shorter than "begin" or "start".  Since those are IMO clearer, I'll change this when I commit it. Thanks!

jeffpc:
FWIW, I'm changing one of the commit messages to reflect that it's base64 encoding specific:

```
lib: Preallocate output space for base64 encoding
```

src/lib/base64.c
compmaniak: Sure! Personally I prefer short names with equal length such as 'beg', 'end', 'cur', 'ptr', etc

compmaniak:
@jeffpc I use github email protection and it displays @users.noreply.github.com instead of my real email. If it looks ugly maybe there is a way to change it.

jeffpc:
It's up to you what you want in the commit.  I just thought I'd ask in case it was an oversight.  Should I use it as-is or do you want something different?

compmaniak:
@jeffpc Ok, you can use it as-is. As long as the project stays on Github the commits are linked to my account. And there will be no problem to find out who write those lines of code.

jeffpc:
Merged.

PR #50, opened 2017-08-31
lib-compression: Fix assert in i_stream_zlib_seek

citrin:
There is a bug in lib-compression which causes panic if I try to use compressed (read-only) mbox as described here: https://wiki2.dovecot.org/Plugins/Zlib

As I can see bug was introduced by this commit:
https://github.com/dovecot/core/commit/6a1110d8757bb72fd90f4fe0857fd3aeaf8920ff (for dovecot 2.2.x)

More info about this problem:
https://www.dovecot.org/pipermail/dovecot/2017-July/108534.html
https://www.dovecot.org/list/dovecot/2017-February/106906.html

If needed, I can describe how this bug can be reproduced.
The fix was tested with gzip, bzip2 and xz compressed mboxes.


cmouse:
merged.

PR #52, opened 2017-09-04
auth: only strip userdb dict prefix if required

manuelm:
I'm not sure this is the correct fix but using FS dictionary as userdb and running `doveadm quota get -A` results in an assert otherwise.

dovecot configuration:
```
uri = fs:posix:prefix=/tmp/dict
iterate_prefix = userdb/
key userdb {
  key = userdb/%u
  format = json
}
```

Adding an additional `auth_request_log_debug`-call to `userdb_dict_get_user` prints:
```
dovecot: auth-worker(pid): Debug: Loading modules from directory: /usr/lib64/dovecot/auth
dovecot: auth-worker(pid): Debug: Module loaded: /usr/lib64/dovecot/auth/lib20_auth_var_expand_crypt.so
dovecot: auth-worker(pid): Debug: dict(,): iterate: prefix=shared/userdb/
dovecot: auth-worker(pid): Debug: dict(,): iterate: key=testuser prefix=shared/userdb/
```

### DIFF
--- a/src/lib-dict-extra/dict-fs.c
+++ b/src/lib-dict-extra/dict-fs.c
@@ -168,12 +168,13 @@ static bool fs_dict_iterate(struct dict_iterate_context *ctx,
 		iter->fs_iter = fs_iter_init(dict->fs, path, 0);
 		return fs_dict_iterate(ctx, key_r, value_r);
 	}
+	path = t_strconcat(iter->paths[iter->path_idx], *key_r, NULL);
 	if ((iter->flags & DICT_ITERATE_FLAG_NO_VALUE) != 0) {
+		*key_r = path;
 		*value_r = NULL;
 		return TRUE;
 	}
 	p_clear(iter->value_pool);
-	path = t_strconcat(iter->paths[iter->path_idx], *key_r, NULL);
 	if ((ret = fs_dict_lookup(ctx->dict, iter->value_pool, path, value_r, &error)) < 0) {
 		/* I/O error */
 		iter->error = i_strdup(error);
@@ -182,6 +183,7 @@ static bool fs_dict_iterate(struct dict_iterate_context *ctx,
 		/* file was just deleted, just skip to next one */
 		return fs_dict_iterate(ctx, key_r, value_r);
 	}
+	*key_r = path;
 	return TRUE;
 }
 


cmouse:
what kind of behaviour do you get without this patch?

manuelm:
An assert/crash.

cmouse:
any chance you could provide logs of that?

manuelm:
Sure. It's the same as above with the assert instead of the added last line:
```
dovecot: auth-worker(pid): Debug: Loading modules from directory: /usr/lib64/dovecot/auth
dovecot: auth-worker(pid): Debug: Module loaded: /usr/lib64/dovecot/auth/lib20_auth_var_expand_crypt.so
dovecot: auth-worker(pid): Debug: dict(,): iterate: prefix=shared/userdb/
dovecot: auth-worker(pid): Panic: file userdb-dict.c: line 123 (userdb_dict_get_user): assertion failed: (strncmp(key, ctx->key_prefix, ctx->key_prefix_len) == 0)
dovecot: auth-worker(pid): Error: Raw backtrace: /usr/lib64/dovecot/libdovecot.so.0(+0x983fa) [0x7f41a8e333fa] -> /usr/lib64/dovecot/libdovecot.so.0(+0x9849f) [0x7f41a8e3349f] -> /usr/lib64/dovecot/libdovecot.so.0(i_fatal+0) [0x7f41a8dcae51] -> dovecot/auth worker: LIST() [0x426f4a] -> dovecot/auth worker: LIST() [0x41befd] -> /usr/lib64/dovecot/libdovecot.so.0(io_loop_call_io+0x52) [0x7f41a8e46e10] -> /usr/lib64/dovecot/libdovecot.so.0(io_loop_handler_run_internal+0x107) [0x7f41a8e48288] -> /usr/lib64/dovecot/libdovecot.so.0(io_loop_handler_run+0x3c) [0x7f41a8e46ea5] -> /usr/lib64/dovecot/libdovecot.so.0(io_loop_run+0x37) [0x7f41a8e47044] -> /usr/lib64/dovecot/libdovecot.so.0(master_service_run+0x13) [0x7f41a8dd48c1] -> dovecot/auth worker: LIST(main+0x300) [0x40d560] -> /lib64/libc.so.6(__libc_start_main+0xf0) [0x7f41a83d9710] -> dovecot/auth worker: LIST(_start+0x29) [0x40d7b9]
dovecot: auth: Error: auth worker: Aborted LIST request for *: Worker process died unexpectedly
dovecot: auth-worker(pid): Fatal: master: service(auth-worker): child pid killed with signal 6 (core dumps disabled)
```

.. which is obvious as you can't strip `shared/userdb` from `testuser`.

cmouse:
well, the thing is that all keys should have full name, that is, shared/userdb/testuser.

cmouse:
or maybe in this case it's enough to create folder userdb/ and put the file there.

manuelm:
> well, the thing is that all keys should have full name, that is, shared/userdb/testuser.

Yeah, I guessed but wasn't completely sure.

> or maybe in this case it's enough to create folder userdb/ and put the file there.

Not sure I understand. The file `/tmp/dict/userdb/testuser` does exist.

manuelm:
@cmouse I just took another look. Latest commit makes sure the key always has the full name.

If you like this approach better I can remove the first or you cherry-pick the second commit.

manuelm:
@cmouse any updates? As already said both commits in this PR fixes the problem. We're with the latter in production.

cmouse:
Hi, sorry this slipped my mind, I'll take a look

cmouse:
The second commit seems to have wrong commit message.

manuelm:
No. Both commits fix the same bug with different approaches/at different locations. You tell me which commit you better like and I'll remove the other. See https://github.com/dovecot/core/pull/52#issuecomment-327159996

We're running with https://github.com/dovecot/core/pull/52/commits/82a8af452eab504d9b8197b2e2715c4527ee51cf

cmouse:
The second commit should say

dict-fs: instead of auth:

manuelm:
Ah yeah, right.

manuelm:
amended

sirainen:
It would be nice to have the important pieces of these explanations here in the commit message(s). Especially that it fixes:

Panic: file userdb-dict.c: line 123 (userdb_dict_get_user): assertion failed: (strncmp(key, ctx->key_prefix, ctx->key_prefix_len) == 0)

manuelm:
Hope it's ok now. Also removed the other commit as it was wrong anyway.

sirainen:
merged

PR #53, opened 2017-09-04
auth: cache: don't log password mismatch twice

manuelm:
If auth cache is enabled and the last auth was successful dovecot assumes the password has been changed and invalidates the cache which results in logging the same password mismatch twice.
This also applies to expired negative cache entries.

Background: We have `auth_verbose` enabled to provide login logging/tracking also for password mismatches.


cmouse:
Hi! Thank you for your contribution, we are currently reviewing this request.

sirainen:
merged

PR #54, opened 2017-09-11
doveadm: let the doveadm-server do the iterating

manuelm:
We recently started issuing more commands over `doveadm-server` instead of manipulating the mailbox ourself and I noticed that `doveadm-server` starts slowing down if a command is issued for more users. Main reason for this is `doveadm` is sending a single request for every user and the per-request overhead of `doveadm-server` is high.

For example fetching the current quota of 5000 mailboxes (`quota = maildir`) takes about 1.2 seconds with `doveadm` as worker vs. about 9 seconds with `doveadm-server`.

Since `doveadm` already passes the commandline arguments over to the server I modified `doveadm` so iterating happens on server-side (62158ccd46737557a3704965597aad2256cd4e54). e.g. For `doveadm quota get -S /../dovadm-server -A` or `doveadm quota get -S /../dovadm-server -u m*lbox`  the users are now iterated on server-side. Since this requires fetching the userlist from `auth-userdb` we have to make sure `doveadm-server` has the required privileges to do so (3a55b5e8a21d8c0b588738599dae9207190fdeee).

Additional I implemented support for passing multiple users to the server (8340aad647b7047c46a23d9617f0a509583e0d61). e.g  `echo -e "mailbox1\nmailbox2\nmailbox3" | doveadm quota get -S /../dovadm-server -F -` only fetches quota for these three mailboxes.

I've split the change in a couple of commits so you can better track the intent.

### DIFF
--- a/src/doveadm/doveadm-mail.c
+++ b/src/doveadm/doveadm-mail.c
@@ -8,6 +8,7 @@
 #include "istream-dot.h"
 #include "istream-seekable.h"
 #include "str.h"
+#include "strescape.h"
 #include "unichar.h"
 #include "module-dir.h"
 #include "wildcard-match.h"
@@ -481,7 +482,7 @@ doveadm_mail_all_users(struct doveadm_mail_cmd_context *ctx,
 
 	ctx->v.init(ctx, ctx->args);
 
-	if (ctx->users_list_input == NULL)
+	if (ctx->users_list == NULL)
 		mail_storage_service_all_init_mask(ctx->storage_service,
 			wildcard_user != NULL ? wildcard_user : "");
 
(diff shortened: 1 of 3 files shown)


src/doveadm/doveadm-mail.c
```diff
@@ -1021,6 +1021,17 @@ doveadm_cmd_ver2_to_mail_cmd_wrapper(struct doveadm_cmd_context *cctx)
 					wildcard_user = arg->value.v_string;
 					mctx->cur_username = NULL;
 				}
+			} else if (strchr(arg->value.v_string, ',') != NULL) {
+				mctx->add_username_header = TRUE;
+				if (!is_client) {
+					char *users = p_str_replace(mctx->pool,
```
cmouse: this should be done using `t_strsplit_tabescaped` and counterpart on sending should use `str_append_tabescaped` and data should be tab separated. you might need to enclose it twice.

src/doveadm/doveadm-mail.c
```diff
@@ -478,8 +478,9 @@ doveadm_mail_all_users(struct doveadm_mail_cmd_context *ctx,
 
 	ctx->v.init(ctx, ctx->args);
 
-	mail_storage_service_all_init_mask(ctx->storage_service,
-		wildcard_user != NULL ? wildcard_user : "");
+	if (ctx->users_list_input == NULL)
```
cmouse: see previous comment, you can just store ARRAY of users here, if necessary.

src/doveadm/doveadm-mail.c
manuelm: Not sure I understand. This is already the server/worker side. I'm reusing the same mechanism the `-F` commandline argument is using which expects the data to be LF terminated.

The wire data is like this:
```
C: \tsomeuser\tquota get\t-u\tmailbox1,mailbox3\n
S: mailbox1\tUser quota\t[...]\tmailbox3\tUser quota\t[...]\n+\n
```

`t_strsplit_tabescaped` splits per tab but I have commas.

cmouse: there is no reason why the username cannot contain tabs so it needs to be tabescaped. username can also contain a comma, so it cannot use comma as separator, it should use tab as separator. so what you need to do is to make the list with tab escaped values, separated by tabs, and run the list again over str_tabescaped. then you need to de-escape it twice on the receiving side.

manuelm: Alright. Will see what I can come up with.

> username can also contain a comma, so it cannot use comma as separator

Yeah, I know. But it was the easiest way forward and wanted to see if you are interested at all. Maybe I should ditch the `-u` argument and use `-F\t<list>` on the wire. Sending over a path to a local file isn't of much use anyway.

manuelm:
@cmouse please take a look

cmouse:
Can you please rebase your merge request? Sorry for delay.

manuelm:
rebased

sirainen:
This seems to break doveadm proxying. For example if you have passwd-file:

loginuser1:{plain}pass::::::
loginuser2:{plain}pass::::::
loginuser3:{plain}pass::::::
proxyuser1:{plain}pass::::::proxy=y host=127.0.0.1 destuser=loginuser1
proxyuser2:{plain}pass::::::proxy=y host=127.0.0.1 destuser=loginuser2
proxyuser3:{plain}pass::::::proxy=y host=127.0.0.1 destuser=loginuser3

Performing these:

doveadm -f tab fetch -u 'loginuser*' 'uid text' mailbox inbox
doveadm -f tab fetch -u 'proxyuser*' 'uid text' mailbox inbox
doveadm -f tab fetch -u '*' 'uid text' mailbox inbox

works currently, but after this change it seems to forget the destuser and goes to a loop where it connects to itself.

PR #55, opened 2017-10-10
Added "LL" to the hex literals that needed it

grumpybozo:
On 32-bit platforms with older compilers (e.g. gcc 4.2 on MacOS 10.6 running on a 1st-gen Core Duo) a 'long' is 4 bytes and the compiler does not automatically use a 'long long' when needed, but instead generates an error. e.g.:

```
   libtool: compile:  /usr/bin/g++-4.2 -DHAVE_CONFIG_H -I. -I../../.. -I../../../src/lib -I../../../src/lib-mail -I../../../src/lib-index -I../../../src/lib-storage -I../../../src/plugins/fts -I../../../src/doveadm -I/opt/local/include/openssl -I/opt/local/include -I/opt/local/include/CLucene/ext -pipe -Os -arch i386 -D__STDC_LIMIT_MACROS -MT lucene-wrapper.lo -MD -MP -MF .deps/lucene-wrapper.Tpo -c lucene-wrapper.cc  -fno-common -DPIC -o .libs/lucene-wrapper.o
In file included from ../../../src/lib/lib.h:33,
                from lucene-wrapper.cc:4:
    ../../../src/lib/byteorder.h:94: error: integer constant is too large for ‘long’ type
    ../../../src/lib/byteorder.h:95: error: integer constant is too large for ‘long’ type
    ../../../src/lib/byteorder.h:96: error: integer constant is too large for ‘long’ type
    ../../../src/lib/byteorder.h:97: error: integer constant is too large for ‘long’ type
    make[4]: *** [lucene-wrapper.lo] Error 1
```
Adding the 'LL' to the end of the 16-digit hex literals that are used to test the structure of 64-bit integers fixes this and avoids any problem which could arise from the compiler using a 32-bit type for those literals that could fit in 32 bites.


sirainen:
Wouldn't ULL be more correct?

jeffpc:
Yeah, they are unsigned. I’ll take care of this PR in the morning.

Jeff.

jeffpc:
Huh... very odd that I haven't see it during my 32-bit builds.  Maybe gcc 5+ are smart enough to do the right thing.  Anyway, thanks for the patch. I'm going to let our CI do its thing, etc.

grumpybozo:
Yes, later C/C++ standards (and hence compilers) follow a specified set of rules to select a suitable integer type for a literal value based on its value and notation rather than requiring an explicit type suffix to get anything bigger than a long. See the details for [C++](http://en.cppreference.com/w/cpp/language/integer_literal#The_type_of_the_literal) and [C](http://en.cppreference.com/w/c/language/integer_constant#The_type_of_the_integer_constant)

sirainen:
merged

PR #56, opened 2017-10-28
Argon2 fixes

jedisct1:
The current `password-scheme-sodium` module mixes constants from a specific algorithm (Argon2i) with algorithm-independent functions.

`crypto_pwhash_str()` is not a synonym for Argon2i. It stands for "the recommended algorithm supported by the libsodium version installed on the system". Which is is not Argon2i any more on current versions.
And `crypto_pwhash_str_verify()` can verify any hash produced by any algorithm ever used by `pwhash_str()`.

This PR explicitly calls that Argon2i version of `pwhash_str()` when the Argon2i parameters are used.
It also adds support for Argon2id. For consistency, the initial scheme was thus renamed `ARGON2I`.

This PR also calls `sodium_init()`. Without proper initialization, only generic, slow implementations will be used. Doing so makes a huge difference, especially on recent Intel CPUs with AVX512.

### DIFF
--- a/src/auth/password-scheme-sodium.c
+++ b/src/auth/password-scheme-sodium.c
@@ -5,7 +5,7 @@
 #include <sodium.h>
 
 static void
-generate_argon2(const char *plaintext, const struct password_generate_params *params,
+generate_argon2i(const char *plaintext, const struct password_generate_params *params,
 		const unsigned char **raw_password_r, size_t *size_r)
 {
 	unsigned long long rounds = params->rounds;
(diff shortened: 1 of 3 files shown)


cmouse:
Hi! thanks, we'll take a look.

src/auth/password-scheme-sodium.c
```diff
@@ -41,12 +67,18 @@ verify_argon2(const char *plaintext, const struct password_generate_params *para
...
+#ifdef crypto_pwhash_ALG_ARGON2ID13
+	{ "ARGON2ID", PW_ENCODING_NONE, 0, verify_argon2,
+	  generate_argon2id },
+#endif
 };
 
 void password_scheme_register_sodium(void)
 {
+	if (sodium_init() < 0)
+		i_fatal("sodium_init failed");
```
cmouse: this should be in one commit

src/auth/test-libpassword.c
```diff
@@ -111,7 +111,7 @@ static void test_password_schemes(void)
 	test_password_scheme("SCRAM-SHA-1", "{SCRAM-SHA-1}4096,GetyLXdBuHzf1FWf8SLz2Q==,NA/OqmF4hhrsrB9KR7po+dliTGM=,QBiURvQaE6H6qYTmeghDHLANBFQ=", "test");
 	test_password_scheme("BLF-CRYPT", "{BLF-CRYPT}$2y$05$11ipvo5dR6CwkzwmhwM26OXgzXwhV2PyPuLV.Qi31ILcRcThQpEiW", "test");
 #ifdef HAVE_LIBSODIUM
-	test_password_scheme("ARGON2", "{ARGON2}$argon2i$v=19$m=32768,t=4,p=1$f2iuP4aUeNMrgu34fhOkkg$1XSZZMWlIs0zmE+snlUIcLADO3GXbA2O/hsQmmc317k", "test");
+	test_password_scheme("ARGON2I", "{ARGON2I}$argon2i$v=19$m=32768,t=4,p=1$f2iuP4aUeNMrgu34fhOkkg$1XSZZMWlIs0zmE+snlUIcLADO3GXbA2O/hsQmmc317k", "test");
```
cmouse: maybe ARGON2 should default to the "whatever is best for the platform" implementation?

jedisct1: Even though there are no plans to change this in a near future, it may not be an Argon2 variant forever.

cmouse: I just wonder if there should be {ARGON2} that defaults to whatever argon wants to spew out.

jedisct1: But a different variants will take a different amount of time to complete given the same CPU work factor (`OPSLIMIT`). So I'd rather use an explicit algorithm name.

cmouse: It is not necessarely a problem. We already have {CRYPT} which eats anything from des, md5, sha and bcrypt.

jedisct1: Maybe call it {SODIUM} instead of {ARGON} then?

cmouse: That would not make sense. I ment that maybe add {ARGON} along with {ARGON2I} and {ARGON2ID}? sodium is a library, not a password scheme.

jedisct1: Yes, but `pwhash_verify()` can verify any algorithm produced by high-level API of the library, not specifically `Argon2*`. In both cases, having `{ARGON}` or `{SODIUM}` or whatever generic thing would allow verification, but not hashing, as the parameters are not compatible between functions.

cmouse: Ok, thank you for clarifying this. Then we proceed with this.

src/auth/password-scheme-sodium.c
cmouse: hi, have you fixed this?

jedisct1: Squashed!

jedisct1: Anything else you would like to change?

cmouse: nope, waiting for internal review

cmouse:
merged with local changes.

PR #57, opened 2017-11-02
Password authentication for redis dict driver

dmz-uk:
Support redis authentication using a standard password= argument in the redis dict driver. This change was proposed last year on the mailing list, I'm resubmitting here as pull request.


cmouse:
merged.

PR #58, opened 2017-11-07
doveadm: fix protocol handshake order

manuelm:
@cmouse while working on rebasing #54 I noticed the doveadm handshake is broken. I tracked this down to the new protocol version 1.1. The expected client version gets send to the client after the client sent the supported protocol. However with tcp or unix sockets the server sends the authentication result beforehand (in `client_connection_tcp_send_auth_handshake`) which results in invalid protocol version detection on client side.

In short the server is sending: `"+\nVERSION\tdoveadm-client\t1\t1\n"` (wrong) vs. `"VERSION\tdoveadm-client\t1\t1\n+\n"` (correct)

### DIFF
--- a/src/doveadm/client-connection-tcp.c
+++ b/src/doveadm/client-connection-tcp.c
@@ -27,6 +27,7 @@ struct client_connection_tcp {
 	struct client_connection conn;
 
 	int fd;
+	int listen_fd;
 	struct io *io;
 	struct istream *input;
 	struct ostream *output;
@@ -41,6 +42,8 @@ struct client_connection_tcp {
 
 static void
 client_connection_tcp_input(struct client_connection_tcp *conn);
+static void
+client_connection_tcp_send_auth_handshake(struct client_connection_tcp *conn);
 static void
 client_connection_tcp_destroy(struct client_connection_tcp **_conn);
 
@@ -502,6 +505,7 @@ client_connection_tcp_input(struct client_connection_tcp *conn)
 			conn->use_multiplex = TRUE;
 		}
 		conn->handshaked = TRUE;
+		client_connection_tcp_send_auth_handshake(conn);
 	}
 	if (!conn->authenticated) {
 		if ((ret = client_connection_tcp_authenticate(conn)) <= 0) {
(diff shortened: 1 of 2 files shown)


sirainen:
This broke skipping authentication with UNIX sockets. I'm not entirely sure why, but the listen_fd check needed to be done earlier. At least if service_count=1 the listen_fd may be closed immediately after the connection is accepted. I rewrote the fix in another way.

The server-connection.c break-removal doesn't seem to cause any real changes. I guess you did it so handshaked will be set to TRUE? It's not really needed, and I renamed the handshaked to just authenticate_sent which describes its functionality better.

manuelm:
> This broke skipping authentication with UNIX sockets. I'm not entirely sure why, but the listen_fd check needed to be done earlier. At least if service_count=1 the listen_fd may be closed immediately after the connection is accepted. I rewrote the fix in another way.

Really? Without my changes I'm not able to do any doveadm commands over unix/tcp sockets. Same for the already released dovecot 2.2.33(.2). I've tested my commit with unix socket + auth over tcp. However I'd love to see your commit.

> The server-connection.c break-removal doesn't seem to cause any real changes. I guess you did it so handshaked will be set to TRUE? It's not really needed, and I renamed the handshaked to just authenticate_sent which describes its functionality better.

Yes. Without removing the `break` every new command fails with `doveadm server sent invalid handshake` as the hanshake+authenticated condition never gets false.

cmouse:
also, what Timo didn't mention is that we made a separate fix, and that has been merged.

sirainen:
But I'm confused now, I don't think anything should have been broken with the old code either. I'll go and test things..

manuelm:
Just tested your commits but still getting:
```
# /root/dovecot/inst/bin/doveadm quota get -S /run/dovecot/doveadm-server -u "foo"
doveadm(foo): Error: doveadm server sent invalid handshake: User quota   STORAGE 3       1228801 0       User quota      MESSAGE 8       -       0
doveadm(foo): Error: /run/dovecot/doveadm-server: Command quota get failed for foo: EOF
Quota name Type Value Limit
```

manuelm:
Yep, it's:
> Without removing the break every new command fails with doveadm server sent invalid handshake as the hanshake+authenticated condition never gets false.

Removing the `break` fixes it for me.

sirainen:
Oh, right, it breaks only with UNIX sockets, and only because the "+" is sent before VERSION.

sirainen:
Oh, and it's still failing because of the handshaked-check. Well, I'll change my other patchset that I thought was just a cleanup to fix this up nicely.

PR #59, opened 2017-11-09
fts-solr: Explicitly ask for XML responses

bratkartoffel:
With recent solr versions (7.0+) the default response
writer changed from XML to JSON. This commit sets the
"wt" query parameter for all requests, so that the right
response type is written.

### DIFF
--- a/src/plugins/fts-solr/fts-backend-solr.c
+++ b/src/plugins/fts-solr/fts-backend-solr.c
@@ -224,7 +224,7 @@ get_last_uid_fallback(struct fts_backend *_backend, struct mailbox *box,
 	int ret = 0;
 
 	str = t_str_new(256);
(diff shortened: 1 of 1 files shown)


cmouse:
and what happens if you have SOLR6?

bratkartoffel:
I'm currently setting up various older versions of solr and verifying it doesn't break stuff.
Which versions should be supported by the fts-solr plugin?

bratkartoffel:
As this PR introduces an additional query parameter it should not break stuff for older versions.
But anyways, I just tested it successfully against 6.6.2 and 6.5.1.

cmouse:
Thank you, we'll put this in internal review

cmouse:
merged.

PR #60, opened 2018-01-13
Always link krb5 when gssapi is requested

thespooler:
Since krb5_free_context is defined in libkrb5, not libgssapi, it must be
linked, even when gssapi is found.
krb5-config --libs gssapi alone will not provide -lkrb5.
Also, krb5-config --version ANYTHING will always return 0 (at least with
Heimdal).


cmouse:
Hi! Thank you for this, we'll look into it.

jeffpc:
LGTM.

FWIW, `krb5-config --version gssapi` returns 1 on Solaris/illumos systems because it doesn't support anything related to gssapi:

```
$ krb5-config --version gssapi
/usr/bin/krb5-config: Unknown option `gssapi' -- use `--help' for usage
$ echo $?
1
```

It is an older, hacked up version of MIT krb5.

cmouse:
merged externally. thank you.

PR #61, opened 2018-01-27
str_escape speedup

compmaniak:
This implementation is 3x faster on -O0 and 2x faster on -O2/-O3 in the worst case when a char to be escaped is the first char in a source string (for example, `"\'foo\'"`). It doesn't affect the perfomance when source string contains no escaped chars.

Code for benchmark:
```
for (int i = 0; i < 2000; ++i) T_BEGIN {
    for (int i = 0; i < 10000; ++i)
        str_escape("\'aaaaaaaaaabbbbbbbbbcccccccccdddddddddd\'");
} T_END;
```

### DIFF
--- a/src/lib/strescape.c
+++ b/src/lib/strescape.c
@@ -6,27 +6,28 @@
 
 const char *str_escape(const char *str)
 {
-	const char *p;
 	string_t *ret;
+	size_t i = 0, j = 0;
 
 	/* see if we need to quote it */
-	for (p = str; *p != '\0'; p++) {
-		if (IS_ESCAPED_CHAR(*p))
+	for (; str[i] != '\0'; ++i) {
+		if (IS_ESCAPED_CHAR(str[i]))
 			break;
 	}
 
-	if (*p == '\0')
+	if (str[i] == '\0')
 		return str;
 
 	/* quote */
-	ret = t_str_new((size_t) (p - str) + 128);
-	str_append_n(ret, str, (size_t) (p - str));
-
-	for (; *p != '\0'; p++) {
-		if (IS_ESCAPED_CHAR(*p))
+	ret = t_str_new(i + 128);
+	for (; str[i] != '\0'; ++i) {
+		if (IS_ESCAPED_CHAR(str[i])) {
+			str_append_n(ret, str + j, i - j);
 			str_append_c(ret, '\\');
-		str_append_c(ret, *p);
+			j = i;
+		}
 	}
+	str_append_n(ret, str + j, i - j);
 	return str_c(ret);
 }
 
--- a/src/lib/strescape.c
+++ b/src/lib/strescape.c
@@ -31,6 +31,20 @@ const char *str_escape(const char *str)
 	return str_c(ret);
 }
 
+void str_append_escaped(string_t *dest, const void *src, size_t src_size)
+{
+	const unsigned char *src_c = src;
+	size_t i = 0, j = 0;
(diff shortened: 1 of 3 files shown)


sirainen:
I think you could get it even faster by using strpbrk(str, "\"\\'") to find the next escaped char instead of a for-loop going one by one. The libc strpbrk() is probably more optimized.

compmaniak:
`strpbrk(str, "\"\'\\")` couldn't be used here. The main point is to reduce the str_append function calling. Appending a block of chars is  faster then sequential appending each char of the block. strpbrk returns null pointer when no more chars found. So we won't know the size of the last block in the case `"bla\\blabla"`

cmouse:
Actually strpbrk works just fine...

````.c
void str_append_escaped(string_t *dest, const void *src, size_t src_size)
{
	const unsigned char *last = src, *ptr;
	const unsigned char *endp = PTR_OFFSET(src, src_size);
	while((ptr = (const unsigned char*)strpbrk((const char*)last, "\\\"'")) != NULL) {
		str_append_c(dest, '\\');
		str_append_n(dest, last, (uintptr_t)(ptr-last)+1);
		last = ptr+1;
	}
	if (last < endp)
		str_append_n(dest, last, (uintptr_t)(endp-last)+1);
}
```

sirainen:
For str_append_escaped() yeah. For strescape() the last call would have to be strlen() which I guess isn't so nice since it potentially goes through the string twice. Still, is that slower or faster?.. I guess slower, but would be nice to verify.

cmouse:
Actually that suggestion was wrong, so I ended up with 
```.c
const char *str_escape(const char *str)
{
        size_t len = strlen(str);
        string_t *dest = t_str_new(len*2);
        str_append_escaped(dest, str, len);
        return str_c(dest);
}

void str_append_escaped(string_t *dest, const void *src, size_t src_size)
{
        const unsigned char *last = src, *ptr, *endp = PTR_OFFSET(src, src_size);
        while((ptr = (const unsigned char*)strpbrk((const char*)last, "\\\"'")) != NULL) {
                str_append_n(dest, last, (uintptr_t)(ptr-last));
                str_append_c(dest, '\\');
                str_append_c(dest, *ptr);
                last = ptr+1;
        }
        if (last < endp)
                str_append_n(dest, last, (uintptr_t)(endp-last)+1);
}
```

cmouse:
Some benchmarks using the provided benchmark code:
 - original code: 1519 nsec / op
 - this pr: 635 nsec / op
 - my proposal: 564 nsec / op

I ran each one 10 times and took average. This pr gives 58% faster performance, and my proposal gives 63% compared to orignal code.

compmaniak:
I think implementing `str_append_escaped` via `strpbrk` is wrong because you shouldn't assume that pointer to source is zero-terminated. Your solution doesn't work in the following example
```
const char *bla = "bla\\bla";
str_append_escaped(str, bla, 5);
test_assert(strcmp(str_c(str), "bla\\\\b") == 0);
```

cmouse:
Fair point. I guess str_append_escaped needs to use your code, it's safer.

compmaniak:
Also implementing `str_escape` via `str_append_escaped` will lose optimization when a source string has no characters to escape and is returned as is.

cmouse:
Yeah, I change this so that your code is in str_append_escaped and mine in str_escape.

cmouse:
unfortunately we need null-safe functions these days, so these become not so fancy. thank you a lot for your effort.

PR #62, opened 2018-02-17
Ignore the authzid used in EXTERNAL auth if we are using client certs

huckabeec:
When setting up client TLS authentication for IMAP, Thunderbird sends whatever is in the username field (that cannot be blank) as the authzid, while Apple Mail sends an anonymous name.  This causes EXTERNAL authentication to break when that authzid doesn't match the client SSL certificate name.  This is a case of broken clients but I think we should ignore the authzid when auth_ssl_username_from_cert has been set in the Dovecot config.


sirainen:
Hmh.. The current behavior allows using certificate as the master user login, and the EXTERNAL mechanism providing the username that will log in. I don't think we want to disable that possibility, even if people currently aren't using it. So, although I hate adding new settings, I think this requires a new one. Maybe auth_external_ignore_authzid.

huckabeec:
Let me know if that is OK

src/auth/auth-settings.c
```diff
@@ -261,6 +261,7 @@ static const struct setting_define auth_setting_defines[] = {
 	DEF(SET_STR, verbose_passwords),
 	DEF(SET_BOOL, ssl_require_client_cert),
 	DEF(SET_BOOL, ssl_username_from_cert),
+	DEF(SET_BOOL, auth_external_ignore_authzid),
```
sirainen: Otherwise ok, but "auth_" prefix is automatically added to all these settings. So you've now actually created auth_auth_external_ignore_authzid setting.

sirainen:
Also please put the final result into a nice single commit.

huckabeec:
See if that is acceptable, hopefully no screw ups this time.

cmouse:
`git rebase -i master` should help you a bit.

huckabeec:
Thanks @cmouse - I may have still cocked that up.

PR #63, opened 2018-02-17
Add spectre variant 2 mitigations

codarrenvelvindron:
GCC version 7.3 has support for retpolin
https://support.google.com/faqs/answer/7625886


sirainen:
Is this really useful for Dovecot though? Seems very unlikely to me that there's a way to do any Spectre attacks with Dovecot, even if untrusted Sieve scripts are allowed.

codarrenvelvindron:
The risk is for compromised software running on the same host to do a spectre-like attack by targetting dovecot. Having retpoline as a hardening flag mitigates the risk. A similar discussion took place in OpenSSH: (https://lists.mindrot.org/pipermail/openssh-unix-dev/2018-February/036656.html), where it was decided that having retpoline would be useful in this scenario.

m4/cc_retpoline.m4
```diff
@@ -0,0 +1,21 @@
...
+    if test $enable_hardening = yes; then
+      case "$host" in
+        *)
+          gl_COMPILER_OPTION_IF([-mfunction-return=thunk -mindirect-branch=thunk], [
+            CFLAGS="$CFLAGS -mfunction-return=thunk -mindirect-branch=thunk"
+            ],
+            [],
+            [AC_LANG_PROGRAM([[
+  #include <pthread.h>
+  __thread unsigned int t_id;
```
sirainen: Why is the test program using pthread?

codarrenvelvindron: Hello,

We have based ourselves on existing implementations.

What would you recommend in the Dovecot context?

sirainen: Oh, I see the same pthread code is used in several other m4 checks already in Dovecot. I wonder why. Dovecot generally doesn't use pthreads, so I was just thinking that pthread usage in these checks is a bit confusing, as if it's important to use them only with pthreads.

cmouse: That's just how they are usually written, it's not about whether pthreads are used or not.

cmouse: It seems empty body is fine too, so `AC_LANG_PROGRAM()`. It generates into `return 0;` only.

codarrenvelvindron:
Some improvements have been made following your review.
Compiles fine.
please see:
[#5190741](https://github.com/dovecot/core/pull/63/commits/5190741b2a1e8989473929ee97e6415eaf2b89fd)

PR #64, opened 2018-02-28
Add auth_external_ignore_authzid setting

huckabeec:
Trying this again. This adds the auth_external_ignore_authzid option so we can ignore the authzid sent during EXTERNAL authentication and using auth_ssl_username_from_cert.  This works around broken clients that may send a username the user can't modify (like Apple Mail) or demand the user enter a username they may not know (like Thunderbird).

### DIFF
--- a/doc/example-config/conf.d/10-auth.conf
+++ b/doc/example-config/conf.d/10-auth.conf
@@ -93,6 +93,14 @@
 # CommonName. 
 #auth_ssl_username_from_cert = no
 
+# Ignore the authzid sent from the client when using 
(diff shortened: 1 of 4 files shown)


huckabeec:
Would like to know if this can be reviewed for acceptance or if it needs more ?

cmouse:
Sorry this has taken way too long to be handled, we'll try to deal with this soon.

sirainen:
Did https://github.com/dovecot/core/commit/17efe7a4041358535ede7d78ecc139ea57ece50c solve this also?

PR #65, opened 2018-03-12
Update schema file to work with Solr 7

alexanderharm:
Adapted the schema file to work with current version of Solr 7.

### DIFF
--- a/doc/solr7-schema.xml
+++ b/doc/solr7-schema.xml
@@ -1,0 +1,152 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+For fts-solr:
+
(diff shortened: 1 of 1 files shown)


sirainen:
It's adding a lot of fields that aren't used by the schema. Would be nicer if it changed only the necessary fields, and even nicer if each such change was explained. I see at least:
 * Solr7 adds the Point types, so those those types need to be changed
 * There are new \_text\_ and \_root\_ fields - I'm not sure what exactly these do and whether they're really needed?

cmouse:
Fixed with https://github.com/dovecot/core/commit/db2b6666d428b4479975f1d571bd3f91ef16a8a8

PR #66, opened 2018-03-14
Add ssl_ca_hash setting

huckabeec:
This adds a ssl_ca_hash option that points to a directory location with hashed trusted certifcate authorities and their CRLs for use when requiring client certificates for login.

In our use case, we have multiple root CAs, multiple intermediate CAs, with thousands of client certificates issued against them.  This makes for rather large CRL lists that the current ssl_ca option tries to pull into memory.  That makes for some bloated processes that can quickly eat up all available memory on the server.

The hashed option lets the process read the CRLs as needed on disk.

PR #67, opened 2018-03-25
update_load_ca

huckabeec:
load_ca changes split out from other pull request #66

### DIFF
--- a/src/lib-ssl-iostream/iostream-openssl-context.c
+++ b/src/lib-ssl-iostream/iostream-openssl-context.c
@@ -10,6 +10,7 @@
 #include <openssl/pem.h>
 #include <openssl/ssl.h>
 #include <openssl/err.h>
+#include <sys/stat.h>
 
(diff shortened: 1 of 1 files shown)


cmouse:
The commit message is bit undescriptive, and still, this contains *two* changes in one commit. First fix how CA's are loaded, then add the ca_hash feature. Also this should compile with openssl 1.0.1, not sure where ca hash directory has been added.

cmouse:
Also, it would be nice if the commit messages would be something like

lib-ssl-iostream: Description of change.

huckabeec:
See if you like this better.  I'll move the other bits into the other pull request.

huckabeec:
Would like to know if this can be reviewed again for acceptance to let me know if it still needs more changes.

huckabeec:
Actually I need to look at this again, I was back porting to 2.2.35 and trying to think on how to split this up better from #70

cmouse:
I think we have already fixed this in 2.3.14.

PR #68, opened 2018-03-26
Make arc4random check more precise

grumpybozo:
Code uses  arc4random_buf() but configure checks for  arc4random(). FreeBSD <8.0 and MacOS X <10.7 have the latter but not the former.


cmouse:
Hi, why did you close this?

grumpybozo:
On 26 Mar 2018, at 0:33, Aki Tuomi wrote:

> Hi, why did you close this?

It was crufty, with old changes that had already been applied upstream 
and a resync from upstream included in the PR but doing effectively 
nothing.

#69 is a clean version.

PR #69, opened 2018-03-26
Fix arc4random check

grumpybozo:
The configure script checks for arc4random() but the actual code uses arc4random_buf(). This breaks for FreeBSD <8.0 and MacOS X <10.7, which have the former but not the latter. This change (and an autoreconf run) solves the problem at the cost of not using the available ARC4 implementation.

### DIFF
--- a/m4/arc4random.m4
+++ b/m4/arc4random.m4
@@ -13,11 +13,11 @@ AC_DEFUN([DOVECOT_ARC4RANDOM], [
      [--with-libbsd],
      [Use libbsd (default is no)]
   ), [want_libbsd=$withval], [want_libbsd=no])
-  AC_CHECK_FUNC([arc4random], AC_DEFINE([HAVE_ARC4RANDOM], [1], [Define this if you arc4random()]), [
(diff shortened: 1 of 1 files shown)


jeffpc:
Is that a sufficient check?  The code on master uses `arc4random()` `arc4random_buf()` and `arc4random_uniform()`.

grumpybozo:
On 26 Mar 2018, at 7:08 (-0400), jeffpc wrote:

> Is that a sufficient check?  The code on master uses `arc4random()` 
> `arc4random_buf()` and `arc4random_uniform()`.

Yes. Both `arc4random_buf()` and `arc4random_uniform()` were added in 
FreeBSD 8.0, MacOS X 10.7, and Debian 6. The previous set of related 
functions on all 3 platforms was `arc4random()`, `arc4random_stir()`, 
and `arc4random_addrandom()`.


-- 
Bill Cole
bill@scconsult.com or billcole@apache.org
(AKA @grumpybozo and many *@billmail.scconsult.com addresses)
Currently Seeking Steady Work: https://linkedin.com/in/billcole

jeffpc:
Fair enough.

cmouse:
merged externally.

PR #70, opened 2018-03-27
lib-ssl-iostream: Add ssl_ca_hash_dir setting

huckabeec:
Adds ssl_ca_hash_dir setting so a hash of CRLs and CAs can be used instead of one large file which can cause problems if you have many root and intermediate CAs along with large CRLs.

Reworked again, this depends on #67

### DIFF
--- a/doc/example-config/conf.d/10-ssl.conf
+++ b/doc/example-config/conf.d/10-ssl.conf
@@ -23,6 +23,22 @@ ssl_key = </etc/ssl/private/dovecot.pem
 # followed by the matching CRL(s). (e.g. ssl_ca = </etc/ssl/certs/ca.pem)
 #ssl_ca = 
 
+# A directory location with hashed trusted certifcate authorities and their
+# CRL lists. An example bash script to build these hashed using OpenSSL:
+# #!/bin/sh
+# for file in *.pem; do ln -s "$file" "$(openssl x509 -hash -noout -in "$file")".0; done
+# for file in *.crl; do ln -s "$file" "$(openssl crl -hash -noout -in "$file")".r0; done
+#
+# Some platforms may also come with the 'c_rehash' command which can also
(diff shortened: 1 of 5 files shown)


cmouse:
After pondering this through I realize this cannot really work. login processes are chrooted on startup, so they cannot access any external directories.

huckabeec:
How would they reach the CA file then ?  This could use the same process.

cmouse:
It is read into memory and kept there.

huckabeec:
Yeah, now I remember.  That was what started me down this path of making these patches - we have so many root and intermediate CAs that reading them all into memory makes for some huge processes.  Could we create a link to the directory during the chroot process to make the directory available then ?

sirainen:
Directories can't be hard linked, only symlinked and that wouldn't help. With Linux you could have bind mounts, which could allow it to work. It would be a bit annoying to configure though.

We've been thinking about some kind of a "ssl-worker" process, which could be handling the SSL certificates. Ideally it would be the one completely parsing the certificates and checking their validity. These kind of features would be easier to implement there, since it wouldn't need to be chrooted. I'm not sure how easy that would be to implement though, or if it would be too inefficient.

huckabeec:
Well if this is mainly used for tls validation, would it just be easier and move to implementing an ocsp type daemon instead ?  It would handle the heavy lifting of handling the CA chains and expose a local OCSP responder for dovecot to use for checks.

cmouse:
That would be an option, yes

huckabeec:
OK, I'll make a new pull and work on something in that direction, thanks.

sirainen:
@stephanbosch has written quite a lot of OCSP server code already a few years ago. But it got somewhat stuck. Anyway that would be a good starting point if you want to do something related to it. We could make it public somewhere I think.

PR #71, opened 2018-04-04
fix login via PAM

tessus:
After upgrading to 2.3, PAM authentication stopped working:

Error in system's security log:
```
PAM audit_log_acct_message() failed: Operation not permitted
```
Error in `dovecot.log`:
```
auth-worker(*REMOVED*): Info: pam(*REMOVED*): pam_authenticate() failed: System error
```

Removing `NoNewPrivileges=true` and adding `CAP_AUDIT_WRITE` to `CapabilityBoundingSet` fixes this error.

It seems the option `NoNewPrivileges=true` is not only a problem for apparmor, but also for PAM.

### DIFF
--- a/dovecot.service.in
+++ b/dovecot.service.in
@@ -20,9 +20,7 @@ PrivateTmp=true
 NonBlocking=yes
 ProtectSystem=full
 PrivateDevices=true
-# disable this if you want to use apparmor plugin
-NoNewPrivileges=true
-CapabilityBoundingSet=CAP_CHOWN CAP_DAC_OVERRIDE CAP_IPC_LOCK CAP_KILL CAP_NET_BIND_SERVICE CAP_SETGID CAP_SETUID CAP_SYS_CHROOT CAP_SYS_RESOURCE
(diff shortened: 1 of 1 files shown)


jleroy:
Hi @tessus. I can confirm that your patch has fixed the issue on my servers.

tessus:
@jleroy I'm not sure, if this PR will ever be merged though.

I think @cmouse mentioned on the mailing lists that they might remove `NoNewPrivileges` and `CapabilityBoundingSet`. I haven't heard back yet in the PR. I'm more than happy to change it and remove both keywords.

cmouse:
Hi! @tessus is right, we'll likely drop those both, and we have a patch for that already. But thank you very much for your effort and sorry for the communications breakdown.

cpaelzer:
Hi,
Sorry to zombie-rez this PR, but I have ran into the same when checking out 2.3.1 and found dovecot.service.in still unchanged up to today.
Unfortunately by the subject titles nothing in [1][2] indicated a thread that I could read myself into.

What is the ETA on such a change, or if it will be a totally different change, what is the thread I'd want to read to know how/when it will be handled?

[1]: https://www.dovecot.org/pipermail/dovecot/2018-May/thread.html
[2]: https://www.dovecot.org/pipermail/dovecot/2018-April/thread.html

sirainen:
Merged now to master. Will be in v2.3.2 release.

PR #76, opened 2018-04-16
lib-sql: pgsql: Added correct notice processor

zeil:
This change adds pgsql prefix to notice and warning messages from postgres server (see https://www.postgresql.org/docs/current/static/libpq-notice-processing.html). Without this we do not understand from which server those messages were received.


cmouse:
Hi! Thank you for your merge request. We'll take a look at this.

sirainen:
Merged in e050f7412b3d1eeb0ca6270e625db1c6aba2f9df

PR #78, opened 2018-04-17
fix: doveadm-who/kick out of order response

bjne:
doveadm-who and kick did not properly flush protocol output
after doveadm_print(), this caused out-of-order response to
clients, example:

C: <tab><tab>who<crlf>
S: +<crlf>
S: username<tab>[...]<missing_crlf>

correct response should have been
S: <username><tab>[...]<crlf>
S: +<crlf>

fixed by adding doveadm_print_flush() calls.


bjne:
What is happening here?

bjne:
What is happening here?

sirainen:
merged to master

PR #79, opened 2018-04-17
Modify NFS attribute caching mechanism to preserve special permission bits

shaun-a-johnson:
Observed with Netapp NFS mounted filesystems that the 'chown' call
that is made in nfs-workarounds.c function nfs_flush_chown_uid results in file mode bits S_ISUID and S_ISGID being dropped. This change simply adds a chmod call back to the original mode to mitigate this.


cmouse:
Thank you for your merge request, we'll take a look at this.

sirainen:
Those calls aren't supposed to actually change anything in the filesystem. They're supposed to flush the caches, not perform any NFS IO.. If it does, that's rather bad already. Even worse when it actually changes the file. In this case changing the permissions back has already created a race condition when another process may temporarily see the wrong permission and fail.

I'd rather just get rid of the mail_nfs_* settings entirely. They're not working properly anyway, and more kludges to them just make them worse..

sirainen:
Maybe using chmod() only, without chown(), would work to flush the cache. Looks like I still have the test program: http://iki.fi/tss/nfstest.c which could be used to check what kind of operations flush the NFS caches.

shaun-a-johnson:
The chown man pages tend to vary on the topic, but I believe the actual change of S_UID/S_GID bits is a result of:

> When the owner or group of an executable file are changed by an unprivileged user the S_ISUID and S_ISGID mode bits are cleared.  POSIX does not specify whether this  also  should happen  when  root does the chown(); the Linux behavior depends on the kernel version.  In case of a non-group-executable file (i.e., one for which the S_IXGRP bit is not set) the S_ISGID bit indicates mandatory locking, and is not cleared by a chown().

I ran the provided test utility and based on the results I am inclined to agree that swapping out chown call to chmod instead should perform the same operation.

I have tested with a brief code update and confirmed that changing from chown to chmod has the desired results as best as I can tell.

The only questions I might have with using chmod in place of chown is the comments at the head of the nfs-workarounds.c file:

> As long as we're not root the file's owner can't change accidentally. If would be possible to also use chmod()/fchmod(), but that's riskier since it could actually cause an unwanted change.

I'm not sure if the context of this comment was more worrying about changing permissions to something undesired if running as the 'root' user or if there was a deeper risk?

Personally I think that setting the permissions to the same value fetched  from the earlier 'stat' call would be safe - but your input would be appreciated.

sirainen:
I was probably mainly thinking about race conditions if another process was changing the file permissions at the same time. That shouldn't really happen though, because Dovecot doesn't change permissions for any files after they have already been created. But a guaranteed-no-op-chown() would have been safer because it couldn't change anything even theoretically. Except, as you found, when it drops the setuid-bit from the file permissions..

shaun-a-johnson:
I have made an update now to use 'chmod' instead of chown/chmod approach.  Please note that I have limited this in scope to 'linux' only (using assumption of symbol ATTRCACHE_FLUSH_CHOWN_UID_1 being indicative of either FreeBSD or Sun - but not linux) as I am uncertain if the same issue would exist with Netapp mounted on Sun or FreeBSD...  I'd rather not fix it if it ain't broken essentially.   
This patch is confirmed as working in a production cluster environment.

PR #82, opened 2018-05-06
Close stderr along with stdout before daemonizing.

psagers:
This has been causing problems with SaltStack on FreeBSD:

https://github.com/saltstack/salt/issues/44848
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=228003

### DIFF
--- a/src/master/main.c
+++ b/src/master/main.c
@@ -861,6 +861,8 @@ int main(int argc, char *argv[])
 		i_fatal("dup2(dev_null_fd) failed: %m");
 	if (!foreground && dup2(dev_null_fd, STDOUT_FILENO) < 0)
 		i_fatal("dup2(dev_null_fd) failed: %m");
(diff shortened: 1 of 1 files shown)


sirainen:
Well, this is a feature that others rely on.
Especially running with daemontools it wants Dovecot to log everything to stderr.

psagers:
Does daemontools allow dovecot to daemonize itself, but also capture stderr? I was under the impression that a supervisor should take responsibility for neither or both, but trying to mix and match was a recipe for sadness. I could certainly be mistaken about that.

The FreeBSD port can work around this by using daemon(8) in the rc script to redirect the standard descriptors. dovecot may want to at least consider an option to fully detatch.

fzipi:
This was fixed on 10/01.

New versions are mail/dovecot 2.3.3, mail/dovecot-pigeonhole  0.5.3.

So I think this can be closed.

PR #83, opened 2018-05-28
lib-storage: reintroduce mail_index_set_fsync_mode

fabianfreyer:
I'm not sure if this is the right way to propose this: This fix is already in master as bf277f94101249133349c0560dad0463c821a2ef, but it would be really great if we could get this into the next 2.3 release.


cmouse:
It is part of next release

fabianfreyer:
Thanks!

PR #85, opened 2018-06-07
passdb-sql: clean usunsed password in memory

miecio45:
On line [#109](https://github.com/dovecot/core/blob/2a3b7083ce4e62a8bd836f9983b223e98e9bc157/src/auth/passdb-sql.c#L109) we use strdup to copy _auth_request->passdb_password_ to _char * password_ variable. This variable is never directly cleaned in memory. I think that it would be good to add some _safe_memset_. I think that it's not breaking anything, but I need Your opinion.


cmouse:
Thank you for your pull request, we'll look into it.

cmouse:
Merged as 59b6c4cdf15d3d6fad68e6a37c268aa52b83eaf7

PR #86, opened 2018-06-21
Support for Capability items to be 'dynamic'

shaun-a-johnson:
This patch converts the previous string-literal representation of the capabilities to a dynamic list of 'capability objects' each reflecting the capability word they intend to add along with flags that represent when the capability word is to be displayed (eg: varying states). 

Upon a request for the capability string the list of capability objects is iterated and a capability string is dynamically built by appending each capability word if the conditions described in the flags of the capability object are met.  

This patch provides full backwards compatibility for existing plugins that might have previously used client_add_capability as well as honors (and converts) existing string literal capability words into equivalent additions to the new list of capability objects.

Some examples of circumstances that can be specified when adding
new capabilities are:

    - Only show at pre-login or post-login
    - Only show if any TLS/SSL is active/inactive
    - Only show if plaintext auth login is enabled

The default operation is to show the capabaility word if at least
one of the circumstances is met, an additional flag provides the
ability to only show the capability word if exactly all of the
circumstances are met.

The API for adding capabilities now simply wraps the API for adding
entries to the capability list with the 'show always' flag set.

Retrieving the capability string is now a loop that iterates the list
of capability objects and evaluates whether each capability should be
added to the capability string based on the state of the current IMAP
session.

### DIFF
--- a/src/imap-login/imap-login-client.c
+++ b/src/imap-login/imap-login-client.c
@@ -20,6 +20,7 @@
 #include "imap-quote.h"
 #include "imap-login-commands.h"
 #include "imap-login-settings.h"
+#include "imap-capability-list.h"
 
 #if LOGIN_MAX_INBUF_SIZE < 1024+2
 #  error LOGIN_MAX_INBUF_SIZE too short to fit all ID command parameters
@@ -92,31 +93,31 @@ static const char *get_capability(struct client *client)
 {
 	struct imap_client *imap_client = (struct imap_client *)client;
 	string_t *cap_str = t_str_new(256);
-	bool explicit_capability = FALSE;
 
-	if (*imap_client->set->imap_capability == '\0')
-		str_append(cap_str, CAPABILITY_BANNER_STRING);
-	else if (*imap_client->set->imap_capability != '+') {
-		explicit_capability = TRUE;
-		str_append(cap_str, imap_client->set->imap_capability);
-	} else {
-		str_append(cap_str, CAPABILITY_BANNER_STRING);
-		str_append_c(cap_str, ' ');
-		str_append(cap_str, imap_client->set->imap_capability + 1);
-	}
+	/* imap-login is preauth by definition */
+	enum imap_capability_visibility visibility = IMAP_CAP_VISIBILITY_PREAUTH;
 
-	if (!explicit_capability) {
-		if (imap_client->set->imap_literal_minus)
-			str_append(cap_str, " LITERAL-");
-		else
-			str_append(cap_str, " LITERAL+");
-	}
+	/* is tls/ssl enabled? */
+	if (client->tls)
+		visibility |= IMAP_CAP_VISIBILITY_TLS_ACTIVE;
+	else
+		visibility |= IMAP_CAP_VISIBILITY_TLS_INACTIVE;
 
-	if (client_is_tls_enabled(client) && !client->tls)
-		str_append(cap_str, " STARTTLS");
+	/* is login cmd disabled? */
 	if (is_login_cmd_disabled(client))
-		str_append(cap_str, " LOGINDISABLED");
+		visibility |= IMAP_CAP_VISIBILITY_NO_LOGIN;
 
+	/* Are we secured? (localhost? tls? etc) */
+	if (imap_client->common.secured)
+		visibility |= IMAP_CAP_VISIBILITY_SECURE;
+	else
+		visibility |= IMAP_CAP_VISIBILITY_INSECURE;
+
+	/* build capability string based on IMAP_CAP_VISIBILITY_ flags */
+	imap_capability_list_get_capability(imap_client->capability_list,
+					    cap_str, visibility);
+
+	/* grab the AUTH= capabilities from auth server */
 	client_authenticate_get_capabilities(client, cap_str);
 	return str_c(cap_str);
 }
@@ -374,6 +375,7 @@ static struct client *imap_client_alloc(pool_t pool)
 static void imap_client_create(struct client *client, void **other_sets)
 {
 	struct imap_client *imap_client = (struct imap_client *)client;
+	bool explicit_capability = FALSE;
 
 	imap_client->set = other_sets[0];
 	imap_client->parser =
@@ -382,6 +384,51 @@ static void imap_client_create(struct client *client, void **other_sets)
 				   IMAP_LOGIN_MAX_LINE_LENGTH);
 	if (imap_client->set->imap_literal_minus)
 		imap_parser_enable_literal_minus(imap_client->parser);
+
+	/* create our capability list from CAPABILITY_BANNER_STRING */
+	imap_client->capability_list =
+		 imap_capability_list_create(NULL);
+
+	if (*imap_client->set->imap_capability == '\0')
+		imap_capability_list_append_string(imap_client->capability_list,
+						   CAPABILITY_BANNER_STRING);
+	else if (*imap_client->set->imap_capability != '+') {
+		imap_capability_list_append_string(imap_client->capability_list,
+						   imap_client->set->imap_capability);
+		explicit_capability = TRUE;
+	} else {
+		/* add the capability banner string to the cap list */
+		imap_capability_list_append_string(imap_client->capability_list,
+						   CAPABILITY_BANNER_STRING);
+		/* add everything after the plus to our cap list */
+		imap_capability_list_append_string(imap_client->capability_list,
+						   imap_client->set->imap_capability + 1);
+	}
+
+	/* Add the LITERAL+/- if the capability isn't explicitly set */
+	if (!explicit_capability) {
+		if (imap_client->set->imap_literal_minus)
+			imap_capability_list_add(imap_client->capability_list, "LITERAL-",
+						 IMAP_CAP_VISIBILITY_ALWAYS);
+		else
+			imap_capability_list_add(imap_client->capability_list, "LITERAL+",
+						 IMAP_CAP_VISIBILITY_ALWAYS);
+	}
+
+	/* add STARTTLS to cap list if it is needed */
+	if (client_is_tls_enabled(client)) {
+		imap_capability_list_add(imap_client->capability_list, "STARTTLS",
+					 IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL |
+					 IMAP_CAP_VISIBILITY_PREAUTH |
+					 IMAP_CAP_VISIBILITY_TLS_INACTIVE);
+	}
+
+	/* add LOGINDISABLED to cap list */
+	imap_capability_list_add(imap_client->capability_list, "LOGINDISABLED",
+				 IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL |
+				 IMAP_CAP_VISIBILITY_PREAUTH |
+				 IMAP_CAP_VISIBILITY_NO_LOGIN);
+
 	client->io = io_add_istream(client->input, client_input, client);
 }
 
@@ -389,6 +436,7 @@ static void imap_client_destroy(struct client *client)
 {
 	struct imap_client *imap_client = (struct imap_client *)client;
 
+	imap_capability_list_unref(&imap_client->capability_list);
 	i_free_and_null(imap_client->proxy_backend_capability);
 	imap_parser_unref(&imap_client->parser);
 }
--- a/src/imap/imap-client.c
+++ b/src/imap/imap-client.c
@@ -26,6 +26,7 @@
 #include "imap-notify.h"
 #include "imap-commands.h"
 #include "imap-feature.h"
+#include "imap-capability-list.h"
 
 #include <unistd.h>
 
@@ -157,17 +158,21 @@ struct client *client_create(int fd_in, int fd_out,
 	client->notify_flag_changes = TRUE;
 	p_array_init(&client->enabled_features, client->pool, 8);
 
-	client->capability_string =
-		str_new(client->pool, sizeof(CAPABILITY_STRING)+64);
+	client->capability_string = imap_capability_list_create(NULL);
 
 	if (*set->imap_capability == '\0')
-		str_append(client->capability_string, CAPABILITY_STRING);
+		imap_capability_list_append_string(client->capability_list,
+						   CAPABILITY_STRING);
 	else if (*set->imap_capability != '+') {
-		str_append(client->capability_string, set->imap_capability);
+		imap_capability_list_append_string(client->capability_list,
+						   set->imap_capability);
 	} else {
-		str_append(client->capability_string, CAPABILITY_STRING);
-		str_append_c(client->capability_string, ' ');
-		str_append(client->capability_string, set->imap_capability + 1);
+		/* add the capability banner string to the cap list */
+		imap_capability_list_append_string(client->capability_list,
+						   CAPABILITY_STRING);
+		/* add everything after the plus to our cap list */
+		imap_capability_list_append_string(client->capability_list,
+						   client->set->imap_capability + 1);
 	}
 	if (client->set->imap_literal_minus)
 		client_add_capability(client, "LITERAL-");
@@ -587,8 +592,36 @@ void client_add_capability(struct client *client, const char *capability)
 		/* explicit capability - don't change it */
 		return;
 	}
-	str_append_c(client->capability_string, ' ');
-	str_append(client->capability_string, capability);
+
+	/* add it to our capability list as CAP_ALWAYS */
+	imap_capability_list_add(client->capability_list,
+				 capability, IMAP_CAP_VISIBILITY_ALWAYS);
+}
+
+const char *client_get_capability(struct client *client)
+{
+	string_t *cap_str = t_str_new(256);
+
+	/* imap is postauth by definition */
+	enum imap_capability_visibility visibility = IMAP_CAP_VISIBILITY_POSTAUTH;
+
+	/* is the client secured by means of ssl? */
+	if (client->ssl_secured)
+		visibility |= IMAP_CAP_VISIBILITY_TLS_ACTIVE;
+	else
+		visibility |= IMAP_CAP_VISIBILITY_TLS_INACTIVE;
+
+	/* Are we secured? (localhost? tls? etc) */
+	if (client->secured)
+		visibility |= IMAP_CAP_VISIBILITY_SECURE;
+	else
+		visibility |= IMAP_CAP_VISIBILITY_INSECURE;
+
+	/* build capability string based on IMAP_CAP_VISIBILITY_ flags */
+	imap_capability_list_get_capability(client->capability_list,
+					    cap_str, visibility);
+
+	return str_c(cap_str);
 }
 
 void client_send_line(struct client *client, const char *data)
--- a/src/lib-imap/imap-capability-list.c
+++ b/src/lib-imap/imap-capability-list.c
@@ -1,0 +1,143 @@
+#include "imap-capability-list.h"
+#include "array.h"
+#include "str.h"
+
+struct imap_capability_list *
+imap_capability_list_create(const char *initial_entries)
+{
+	/* allocate the pool for the capability list */
+	pool_t pool = pool_alloconly_create(MEMPOOL_GROWING"Capability List", 512);
+	/* initialize a capability list object */
+	struct imap_capability_list *
+		capability_list = p_new(pool, struct imap_capability_list, 1);
+	capability_list->refcount = 1;
+	/* initialize the capability list memory pool */
+	capability_list->pool = pool;
+	/* initialize the capability list capability array */
+	p_array_init(&(capability_list->imap_capabilities), pool, 64);
+
+	/* make sure our initial entries are appended */
+	if (initial_entries != NULL)
+		imap_capability_list_append_string(capability_list, initial_entries);
+
+	return capability_list;
+}
+
+void imap_capability_list_ref(struct imap_capability_list *capability_list)
+{
+	i_assert(capability_list->refcount > 0);
+
+	capability_list->refcount++;
+}
+
+void imap_capability_list_unref(struct imap_capability_list **_capability_list)
+{
+        struct imap_capability_list *capability_list = *_capability_list;
+
+	*_capability_list = NULL;
+	i_assert(capability_list->refcount > 0);
+
+	if (--capability_list->refcount > 0)
+		return;
+
+	pool_unref(&capability_list->pool);
+}
+
+void imap_capability_list_add(struct imap_capability_list *capability_list,
+			      const char *name,
+			      enum imap_capability_visibility visibility)
+{
+	/* make sure it isn't already in the list */
+	i_assert(imap_capability_list_find(capability_list, name) == NULL);
+
+	struct imap_capability capability;
+	/* initialize the capability object */
+	capability.name = p_strdup(capability_list->pool, name);
+	capability.visibility = visibility;
+
+	/* append new capability object to array */
+	array_push_back(&capability_list->imap_capabilities, &capability);
+}
+
+void imap_capability_list_remove(struct imap_capability_list *capability_list,
+				 const char *name)
+{
+	/* iterate array of capabilities and find our capability */
+	unsigned int i, count;
+	const struct imap_capability *cap =
+		array_get(&capability_list->imap_capabilities, &count);
+	for (i = 0; i < count; i++) {
+		if (strcasecmp(cap[i].name, name) == 0) {
+			array_delete(&capability_list->imap_capabilities, i, 1);
+			return;
+		}
+	}
+	i_unreached();
+}
+
+void imap_capability_list_append_string(struct imap_capability_list *capability_list,
+					const char *name_list)
+{
+	T_BEGIN {
+		/* split the name list by space */
+		const char **strings = t_strsplit_spaces(name_list, " ");
+		/* iterate each name and append it to list as ALWAYS */
+		while (*strings != NULL) {
+			imap_capability_list_add(capability_list, *strings,
+						 IMAP_CAP_VISIBILITY_ALWAYS);
+			strings++;
+		}
+	} T_END;
+}
+
+static int
+imap_capability_bsearch(const char *name, const struct imap_capability *cap)
+{
+	return strcasecmp(name, cap->name);
+}
+
+struct imap_capability *
+imap_capability_list_find(struct imap_capability_list *capability_list,
+			  const char *name)
+{
+	return array_bsearch(&capability_list->imap_capabilities,
+			     name, imap_capability_bsearch);
+}
+
+void imap_capability_list_get_capability(struct imap_capability_list *capability_list,
+					 string_t *cap_str,
+					 enum imap_capability_visibility visibility)
+{
+	/* iterate capability list and append each capability if conditions met */
+	const struct imap_capability *capp;
+	array_foreach(&capability_list->imap_capabilities, capp) {
+		/* local visibility variable will get filled with all the
+		   visibility flags that matched the current conditions --
+		   then at the end if FLAG_REQUIRE_ALL is present the local
+		   visibility variable is compared against the capability
+		   flags to check for a perfect match */
+		if (capp->visibility) {
+			/* grab the visibility of the current cap into a tmp variable */
+			enum imap_capability_visibility tmp_visibility = capp->visibility;
+			/* strip off the REQUIRE_ALL and any visibility flags that don't match */
+			tmp_visibility &= (visibility & ~IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL);
+
+			bool add = FALSE;
+			/* now check if IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL is present */
+			if ((capp->visibility & IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL) != 0) {
+				/* did we match all? */
+				if ((capp->visibility & ~IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL) == tmp_visibility)
+					add = TRUE;
+			/* no IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL - did we match any visibility flags? */
+			} else if (tmp_visibility > 0)
+				add = TRUE;
+
+			/* should we add this entry? */
+			if (add) {
+				if (str_len(cap_str) != 0)
+					str_append_c(cap_str, ' ');
+				str_append(cap_str, capp->name);
+			}
+		}
+	}
+}
--- a/src/lib-imap/imap-capability-list.h
+++ b/src/lib-imap/imap-capability-list.h
@@ -1,0 +1,76 @@
+#ifndef IMAP_CAPABILITY_LIST_H
+#define IMAP_CAPABILITY_LIST_H
+
+#include "lib.h" /* for BIT(n) */
+
+/* bitflags to represent different conditions of
+   the current imap session */
+enum imap_capability_visibility {
+	IMAP_CAP_VISIBILITY_NEVER               = 0,      /* never show this capability */
+	IMAP_CAP_VISIBILITY_PREAUTH             = BIT(0), /* show this cap before login */
+	IMAP_CAP_VISIBILITY_POSTAUTH            = BIT(1), /* show this cap after login */
+	IMAP_CAP_VISIBILITY_TLS_ACTIVE          = BIT(2), /* show this cap if tls/ssl is active */
+	IMAP_CAP_VISIBILITY_TLS_INACTIVE        = BIT(3), /* show this cap if tls/ssl is inactive */
+	IMAP_CAP_VISIBILITY_NO_LOGIN            = BIT(6), /* show this cap if login is disabled */
+	/* NOTE: connections from localhost are 'secure' regardless of SSL/TLS! */
+	IMAP_CAP_VISIBILITY_SECURE              = BIT(7), /* show this cap if we are secured */
+	IMAP_CAP_VISIBILITY_INSECURE            = BIT(8), /* show this cap if we aren't secured */
+
+	/* IMAP_CAP_VISIBILITY_REQUIRE_ALL indicates that the rest
+	   of the conditions must be met *exactly* to show this
+	   capability, otherwise if *at least* 1 condition is met
+	   the capability will be shown */
+	IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL     = BIT(30),
+
+	IMAP_CAP_VISIBILITY_ALWAYS               = 0x3FFFFFFF /* always show this cap */
+};
+
+/* This structure describes a capability word that appears
+   in the imap and imap-login capability string.
+   The visibility indicates when and under which conditions
+   the capability word should be displayed */
+struct imap_capability {
+	const char *name;
+	enum imap_capability_visibility visibility;
+};
+
+/* define our array type for the imap capability array */
+ARRAY_DEFINE_TYPE(imap_capability_array, struct imap_capability);
+
+/* define the structure that holds the capability pool and capability array */
+struct imap_capability_list {
+	/* reference count */
+	int refcount;
+	/* imap capability list memory pool */
+	pool_t pool;
+	/* array of imap_capability objects */
+	ARRAY_TYPE(imap_capability_array) imap_capabilities;
+};
+
+/* initialize the imap capability list */
+struct imap_capability_list *
+imap_capability_list_create(const char *initial_entries);
+/* increment reference to an imap capability list */
+void imap_capability_list_ref(struct imap_capability_list *capability_list);
+/* decrement reference to an imap capability list */
+void imap_capability_list_unref(struct imap_capability_list **capability_list);
+/* add an entry to the imap capability list */
+void imap_capability_list_add(struct imap_capability_list *capability_list,
+			      const char *name,
+			      enum imap_capability_visibility visibility);
+/* remove an entry from the imap capability list */
+void imap_capability_list_remove(struct imap_capability_list *capability_list,
+				 const char *name);
+/* split a string by space and add each entry to the imap capability list */
+void imap_capability_list_append_string(struct imap_capability_list *capability_list,
+					const char *name_list);
+/* find an imap capability by name */
+struct imap_capability *
+imap_capability_list_find(struct imap_capability_list *capability_list,
+			  const char *name);
+/* append the capabilities to cap_str based on given IMAP_CAP_VISIBILITY_ visibility */
+void imap_capability_list_get_capability(struct imap_capability_list *capability_list,
+					 string_t *cap_str,
+					 enum imap_capability_visibility visibility);
+
+#endif
--- a/src/imap-login/imap-login-client.h
+++ b/src/imap-login/imap-login-client.h
@@ -56,6 +56,7 @@ struct imap_client {
 
 	const struct imap_login_settings *set;
 	struct imap_parser *parser;
+	struct imap_capability_list *capability_list;
 	char *proxy_backend_capability;
 
 	const char *cmd_tag, *cmd_name;
--- a/src/imap/cmd-capability.c
+++ b/src/imap/cmd-capability.c
@@ -7,7 +7,7 @@
 bool cmd_capability(struct client_command_context *cmd)
 {
 	client_send_line(cmd->client, t_strconcat(
-		"* CAPABILITY ", str_c(cmd->client->capability_string), NULL));
+		"* CAPABILITY ", client_get_capability(cmd->client), NULL));
 
 	client_send_tagline(cmd, "OK Capability completed.");
 	return TRUE;
--- a/src/imap/imap-client.h
+++ b/src/imap/imap-client.h
@@ -169,7 +169,7 @@ struct client {
 	struct mail_storage_service_user *service_user;
 	const struct imap_settings *set;
 	const struct smtp_submit_settings *smtp_set;
-	string_t *capability_string;
+	struct imap_capability_list *capability_list;
 	const char *disconnect_reason;
 
         struct mail_user *user;
@@ -287,6 +287,9 @@ void client_disconnect_with_error(struct client *client,
 /* Add the given capability to the CAPABILITY reply. If imap_capability setting
    has an explicit capability, nothing is changed. */
 void client_add_capability(struct client *client, const char *capability);
+
+/* Generate the string of capabilities from the client */
+const char *client_get_capability(struct client *client);
 
 /* Send a line of data to client. */
 void client_send_line(struct client *client, const char *data);
--- a/src/imap/imap-client.h
+++ b/src/imap/imap-client.h
@@ -236,6 +236,8 @@ struct client {
 	bool mailbox_examined:1;
 	bool anvil_sent:1;
 	bool tls_compression:1;
+	bool secured:1;
+	bool ssl_secured:1;
 	bool input_skip_line:1; /* skip all the data until we've
 					   found a new line */
 	bool modseqs_sent_since_sync:1;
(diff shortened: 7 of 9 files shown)


src/imap-login/imap-login-client.c
```diff
@@ -93,31 +94,80 @@ static const char *get_capability(struct client *client)
...
+	/* iterate capability list and append each capability if conditions met */
+	const struct imap_capability *capp;
+	array_foreach(&imap_client->capability_list->imap_capabilities, capp) {
+		/* local flags variable will get filled with all the flags
+		   that matched the current conditions -- then at the end
+		   if FLAG_REQUIRE_ALL is present the local flags variable
+		   is compared against the command flags to check for a
+		   perfect match */
+		uint32_t flags = 0;
+		if ((*capp).flags) {
```
jeffpc: Please use `capp->flags`.

Also, since flags is not a bool, you'll want an explicit check here (`if (capp->flags != 0)`) per https://wiki2.dovecot.org/Design/Code#Boolean_expressions

src/imap-login/imap-login-client.c
```diff
@@ -93,31 +94,80 @@ static const char *get_capability(struct client *client)
...
+					client->tls) {
+				flags |= CAP_SECURE;
+			}
+			/*  are we insecure? */
+			if (((*capp).flags & CAP_INSECURE) &&
+					!client->tls) {
+				flags |= CAP_INSECURE;
+			}
+
+			int add = 0;
```
jeffpc: `bool` not `int`

src/imap/imap-client.c
```diff
@@ -546,12 +580,90 @@ void client_add_capability(struct client *client, const char *capability)
...
+				if (*str_c(cap_str) != '\0') {
+					str_append_c(cap_str, ' ');
+				}
+				str_append(cap_str, (*capp).command);
+			}
+		}
+	}
+
+	/* grab the AUTH= capabilities from auth server */
+	//client_authenticate_get_capabilities(client, cap_str);
```
jeffpc: dead code?

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,129 @@
+#include "imap-capability-list.h"
+
+/* initialize the imap capability list */
+struct imap_capability_list *
+imap_capability_list_create(const char *initial_entries)
+{
+    /* initialize a capability list object */
```
jeffpc: It might be github mis-rendering this, but it looks like the indentation here is all wrong.  It should use tabs for indentation.

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,81 @@
...
+#include <stddef.h>
+
+#include "lib.h"
+#include "array.h"
+
+/* bitflags to represent different conditions of
+   the current imap session */
+
+#define CAP_NEVER       0        /* never show this capability */
+#define CAP_PREAUTH     (1 << 0) /* show this cap before login */
```
jeffpc: Using an enum instead of #defines makes gdb/other debuggers pretty-print the values.

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,81 @@
...
+
+/*  the initial size of the capability list
+    it gets automatically increased but this
+    is the size that we start at */
+#define INITIAL_CAP_LIST_SIZE 5
+/*  The amount to grow the capability list
+    each time it fills up */
+#define CAP_LIST_GROW_AMOUNT  1
+
+/* XXX: This really should be integrated with the imap_commands */
```
jeffpc: Ok.  Is this something that should be done before merging this?  Or does that change the API in some way?

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,81 @@
...
+/* decrement reference to an imap capability list */
+void imap_capability_list_unref(struct imap_capability_list **capability_list);
+/* add an entry to the imap capability list */
+void imap_capability_list_add(struct imap_capability_list *capability_list,
+        const char *command, uint32_t flags);
+/* remove an entry from the imap capability list */
+void imap_capability_list_remove(struct imap_capability_list *capability_list,
+        const char *command);
+/* split a string by space and add each entry to the imap capability list */
+void imap_capability_list_append_string(
```
jeffpc: It is ok to go over 80 columns in dovecot.  It avoids some awkward line breaks given the large number of very long symbol/identifier names.

jeffpc:
The commit history of the PR needs a bit of cleanups as well.

shaun-a-johnson:
I have tended to all of the above comments and adjusted the commit history with git rebase.
Please consider this change for merge with the upstream source, and please let me know if there are any further issues or changes that can be applied.

src/lib-imap/imap-capability-list.h
shaun-a-johnson: This was leftover of some thoughts when creating the original patch, it is no longer relevant and has been changed to a comment describing the respective structure.

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,127 @@
+#include "imap-capability-list.h"
+
+/* initialize the imap capability list */
+struct imap_capability_list *
+imap_capability_list_create(const char *initial_entries)
+{
+	/* initialize a capability list object */
+	struct imap_capability_list *
+		capability_list = i_malloc(sizeof(struct imap_capability_list));
```
sirainen: i_new(struct imap_capability_list)

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,127 @@
...
+	capability_list->refcount = 1;
+	/* initialize the capability list memory pool */
+	capability_list->pool = pool_alloconly_create(MEMPOOL_GROWING"Capability List", 2048);
+	/* initialize the capability list capability array */
+	p_array_init(&(capability_list->imap_capabilities),
+			capability_list->pool,
+			0);
+
+	/* make sure our initial entries are appended */
+	if (initial_entries) {
```
sirainen: initial_entries != NULL (we have a patched clang that gives a warning if boolean expressions are mixed with ints or pointers - it has found quite a lot of accidental bugs)

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,127 @@
...
+	capability_list->refcount++;
+}
+
+/* decrement reference to an imap capability list */
+void imap_capability_list_unref(struct imap_capability_list **capability_list)
+{
+	i_assert((*capability_list)->refcount > 0);
+
+	if (--(*capability_list)->refcount > 0)
+		return;
```
sirainen: Generally in Dovecot deinit functions that take **ptr always set *ptr=NULL regardless of the resulting refcount.

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,127 @@
...
+	i_free(*capability_list);
+	*capability_list = NULL;
+}
+
+/* add an entry to the imap capability list */
+void imap_capability_list_add(struct imap_capability_list *capability_list,
+		const char *command, capability_cond_t flags)
+{
+	if (!capability_list || !command) {
+		return;
```
sirainen: Generally Dovecot functions don't check whether their parameters are NULL. Crash is good enough indication that the code is buggy. In some cases they could be i_assert()s instead. Anyway returning silently most likely just causes a silent bug.

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	/* is it already added? */
+	if (imap_capability_list_find(capability_list, command) != NULL) {
+		return;
```
sirainen: I wonder if double-adding is a bug or an expected situation, i.e. should it silently return or assert-crash?..

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,127 @@
...
+	/* iterate array of capabilities and find our command */
+	unsigned int i, count;
+	const struct imap_capability *cap =
+		array_get(&capability_list->imap_capabilities, &count);
+	for (i = 0; i < count; i++) {
+		if (strcasecmp(cap[i].command, command)) {
+			array_delete(&capability_list->imap_capabilities, i, 1);
+			return;
+		}
+	}
```
sirainen: Related to double-addition: If it wasn't possible to double-add, I'd add i_unreached() here. But if it's ok to double-add, then double-remove shouldn't crash either. But then again if there's no i_unreached() then a typoed command parameter could result in silent failure.

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,127 @@
...
+{
+	if (!capability_list || !command_list) {
+		return;
+	}
+
+	T_BEGIN {
+		/* split the command list by space */
+		const char **strings = t_strsplit_spaces(command_list, " ");
+		/* iterate each command and append it to list as CAP_ALWAYS */
+		while (strings && *strings) {
```
sirainen: while (*strings != NULL) - it's not possible that strings==NULL

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,127 @@
...
+		}
+	} T_END;
+}
+
+/* capability list bsearch comparison callback */
+static int imap_capability_bsearch(const char *command,
+		const struct imap_capability *cap)
+{
+	if (!command || !cap) {
+		return 0;
```
sirainen: I don't think these can ever be NULL

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,83 @@
...
+	CAP_TLSACTIVE   = (1 << 2),  /* show this cap if tls is active */
+	CAP_TLSINACTIVE = (1 << 3),  /* show this cap if tls is inactive */
+	CAP_SSLACTIVE   = (1 << 4),  /* show this cap if ssl is active */
+	CAP_SSLINACTIVE = (1 << 5),  /* show this cap if ssl is inactive */
+	CAP_NO_LOGIN    = (1 << 6),  /* show this cap if login is disabled */
+	/* NOTE: connections from localhost are 'secure' regardless of SSL/TLS! */
+	CAP_SECURE      = (1 << 7),  /* show this cap if we are secured */
+	CAP_INSECURE    = (1 << 8),  /* show this cap if we aren't secured */
+	CAP_ALWAYS      = 0x7FFFFFFF /* always show this cap */
+} capability_cond_t;
```
sirainen: Dovecot typically doesn't use typedefs except for a few of its core library things. Also using CAP_ prefix isn't descriptive enough and could be confused with CAP_* defines in linux/capability.h. I think could be:

enum imap_capability_visiblity {
  IMAP_CAPABILITY_VISIBILITY_...

More verbose, but clearer. Perhaps IMAP_CAP_VISIBILITY_* or maybe there's a better name than "visibility". But still, "imap_" prefix at least and more verbosity would be better.

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,83 @@
...
+/* bitflags to represent different conditions of
+   the current imap session */
+typedef enum {
+	CAP_NEVER       = 0,         /* never show this capability */
+	CAP_PREAUTH     = (1 << 0),  /* show this cap before login */
+	CAP_POSTAUTH    = (1 << 1),  /* show this cap after login */
+	CAP_TLSACTIVE   = (1 << 2),  /* show this cap if tls is active */
+	CAP_TLSINACTIVE = (1 << 3),  /* show this cap if tls is inactive */
+	CAP_SSLACTIVE   = (1 << 4),  /* show this cap if ssl is active */
+	CAP_SSLINACTIVE = (1 << 5),  /* show this cap if ssl is inactive */
```
sirainen: Why is there both TLS and SSL? Also it's wrong to call TLS connection started with STARTTLS command as "TLS" and TLS connection started with the imaps port "SSL". I don't think there's any need to separate them anyway? Simpler to just have SSLACTIVE and SSL_INACTIVE? (Dovecot generally calls all of these "ssl" rather than "tls", although maybe I should have made them all "tls" since the beginning. But too late to change now.)

Also, instead of 1<<n use BIT(n)

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,83 @@
...
+/* initialize the imap capability list */
+struct imap_capability_list *
+imap_capability_list_create(const char *initial_entries);
+/* increment reference to an imap capability list */
+void imap_capability_list_ref(struct imap_capability_list *capability_list);
+/* decrement reference to an imap capability list */
+void imap_capability_list_unref(struct imap_capability_list **capability_list);
+/* add an entry to the imap capability list */
+void imap_capability_list_add(struct imap_capability_list *capability_list,
+		const char *command, capability_cond_t flags);
```
sirainen: Capabilities aren't "command"s. I think the parameter should be either "capability" or "name".

src/imap/imap-client.c
```diff
@@ -552,8 +586,84 @@ void client_add_capability(struct client *client, const char *capability)
...
+	string_t *cap_str = t_str_new(256);
+
+	/* iterate capability list and append each capability if conditions met */
+	const struct imap_capability *capp;
+	array_foreach(&client->capability_list->imap_capabilities, capp) {
+		/* local flags variable will get filled with all the flags
+		   that matched the current conditions -- then at the end
+		   if FLAG_REQUIRE_ALL is present the local flags variable
+		   is compared against the command flags to check for a
+		   perfect match */
```
sirainen: Too much code duplication here. Maybe instead imap_capability_list_get_capabilities() where the caller gives the current CAP_* as parameter? So something like:

```
flags = CAP_POSTAUTH;
if (client->starttls)
  flags |= CAP_TLSACTIVE;
else
  flags |= CAP_TLSINACTIVE;
...
imap_capability_list_get_capabilities(cap_str, flags);

File: src/lib-imap/imap-capability-list.h (hunk 16 above)
Comment (shaun-a-johnson) [reply src/lib-imap/imap-capability-list.h]: Interesting, one of the comments by jeffpc actually asked to switch this to a typedef for the sake of gdb pretty-printing the values of variables. 

I have adjusted it to the enum you described (enum imap_capability_visibility) and I think your suggestion of IMAP_CAP_VISIBILITY_ is much better than the current CAP_ prefix, I have also updated that to the more verbose prefix.

File: src/imap/imap-client.c (hunk 19 above)
Comment (shaun-a-johnson) [reply src/imap/imap-client.c]: Yes, I like this approach much better.

If the imap_capability_list_get_capabilities() function does not distinguish between the meaning of each IMAP_CAP_VISIBILITY_ flag then it could be done in a general form with a for-loop like this:

        uint32_t flag_accumulator = 0;
        uint32_t n = 0;
        if (capp->flags) {
            /* iterate all known IMAP_CAP_VISIBILITY_ flags */
            for (n = 0; n < 9; ++n) {
                if ((capp->flags & BIT(n)) != 0) {
                    flag_accumulator |= n;
                }
            }
            bool add = FALSE;
            /*  now check if CAP_REQUIRE_ALL is present */
            if ((capp->flags & CAP_REQUIRE_ALL) != 0) {
                /*  did we match all? */
                if ((capp->flags & ~CAP_REQUIRE_ALL) == flags) {
                    add = TRUE;
                }
            /*  no CAP_REQUIRE_ALL - did we match any flags? */
            } else if (flags > 0) {
                add = TRUE;
            }
            /*  should we add this entry? */
            if (add) {
                if (*str_c(cap_str) != '\0') {
                    str_append_c(cap_str, ' ');
                }
                str_append(cap_str, capp->command);
            }
        }

As opposed to checking the individual flags and the whether the conditions they require are met:

        /*  is starttls active? */
        if ((capp->flags & CAP_TLSACTIVE) != 0 &&
                        client->starttls) {
                flags |= CAP_TLSACTIVE;
        }

Do you have any reservations about the generalized for-loop approach, for example an issue I have with the for-loop approach is that the 9 is hardcoded into the for-loop being the highest bit-level that IMAP_CAP_VISIBILITY_ flags reach (0 - 8). Do you know if there is a better way to do this without hardcoding the 9?

I often use the counting property of enums to produce a _count entry:

    enum val_enum {
        val_invalid = -1, 

        val_0,
        val_1,
        val_2,

        val_count // always reflects the total number of val_ entries
    };

This allows for checks like:

    if (val <= val_invalid || val >= val_count) {
        // val is outside valid bounds
    }

And iteration in the form of:

    enum val_enum val = val_invalid;
    for (val = val0; val < val_count; ++val) {
        /* iterates all enumerated val_ entries */
    }

But the above approach would need some adjustments to be applied in the context of bitflags and enumerated bitflags.

Of course I don't have to condense into a for-loop, I could just leave the imap_capability_list_get_capabilities() function to manually check for each IMAP_CAP_VISIBILITY_ flag and manually OR the flag into the flag_accumulator the way it was being done previously.

Any suggestions at all would be appreciated!

File: src/lib-imap/imap-capability-list.h (hunk 17 above)
Comment (shaun-a-johnson) [reply src/lib-imap/imap-capability-list.h]: I have adjusted the code to use BIT(n), and I have removed the SSL_ACTIVE/INACTIVE and retained the IMAP_CAP_VISIBILITY_TLS_ACTIVE and IMAP_CAP_VISIBILITY_TLS_INACTIVE.

**Comment by shaun-a-johnson** [comment]
Thank you for reviewing this work Timo, I have tended to all of your suggestions and I have divided the changeset into four logical commits.

If anybody could review this changeset again now that I have pushed the updated branch with all the changes it would be greatly appreciated.

**Comment by shaun-a-johnson** [comment]
Was wondering if there were any additional changes required for this request? 
I fully realize it's a pretty large changeset - if it would be easier to break it down into smaller chunks please let me know and I'd be happy to accomodate.

**Comment by shaun-a-johnson** [comment]
Greetings again  dovecot team,

I know you're all busy with a recent release but I wanted to check in again to see if there is anything that I can do to help expedite this pull request?  Still uncertain if the recent changes made satisfy the changes requested by Timo or if additional updates are still required?

Thank you in advance!

File: src/lib-imap/imap-capability-list.c [hunk 20]
```diff
@@ -0,0 +1,153 @@
...
+/* capability list bsearch comparison callback */
+static int imap_capability_bsearch(const char *name,
+		const struct imap_capability *cap)
+{
+	return strcasecmp(name, cap->name);
+}
+
+/* capability list bsearch 'find' function */
+struct imap_capability *imap_capability_list_find(struct imap_capability_list *capability_list,
+		const char *name)
```
sirainen: Coding style consistency: Better would be:

```
struct imap_capability *
imap_capability_list_find(struct imap_capability_list *capability_list,
			  const char *name)
```

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,153 @@
...
+{
+	return array_bsearch(&capability_list->imap_capabilities,
+			name, imap_capability_bsearch);
```
sirainen: whitespace also

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,85 @@
...
+		const char *name);
+/* split a string by space and add each entry to the imap capability list */
+void imap_capability_list_append_string(struct imap_capability_list *capability_list,
+		const char *name_list);
+/* find an imap capability by name */
+struct imap_capability *imap_capability_list_find(struct imap_capability_list *capability_list,
+		const char *name);
+/* append the capabilities to cap_str based on given IMAP_CAP_VISIBILITY_ flags */
+void imap_capability_list_get_capability(struct imap_capability_list *capability_list, 
+		string_t *cap_str, uint32_t flags);
```
sirainen: `enum imap_capability_visibility flags` would be better, right? And rename flags -> visibility?

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,85 @@
+#ifndef IMAP_CAPABILITY_LIST_H
+#define IMAP_CAPABILITY_LIST_H
+
+#include <stdint.h>
+#include <stddef.h>
```
sirainen: These are already part of lib.h, no need to include

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,85 @@
...
+
+#include "lib.h"
+#include "array.h"
```
sirainen: .c files should include these, not .h files -> remove

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,153 @@
...
+					add = TRUE;
+				}
+			/*  no IMAP_CAP_VISIBILITY_REQUIRE_ALL - did we match any flags? */
+			} else if (tmp_flags > 0) {
+				add = TRUE;
+			}
+
+			/*  should we add this entry? */
+			if (add) {
+				if (*str_c(cap_str) != '\0') {
```
sirainen: str_len(cap_str) != 0

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,153 @@
...
+{
+	/* iterate capability list and append each capability if conditions met */
+	const struct imap_capability *capp;
+	array_foreach(&capability_list->imap_capabilities, capp) {
+		/* local flags variable will get filled with all the flags
+		   that matched the current conditions -- then at the end
+		   if FLAG_REQUIRE_ALL is present the local flags variable
+		   is compared against the capability flags to check for a
+		   perfect match */
+		uint32_t n = 0;
```
sirainen: n is not used

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,153 @@
...
+		if (capp->flags) {
+			/* grab the flags of the current cap into a tmp variable */
+			uint32_t tmp_flags = capp->flags;
```
sirainen: I'd keep this `enum imap_capability_visibility` as well still, even though there's a bunch of bitmask operations..

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,153 @@
...
+			/* strip off the REQUIRE_ALL and any flags that don't match */
+			tmp_flags &= (flags & ~IMAP_CAP_VISIBILITY_REQUIRE_ALL);
+
+			bool add = FALSE;
+			/*  now check if IMAP_CAP_VISIBILITY_REQUIRE_ALL is present */
```
sirainen: here and below these comments have two spaces after /* for some reason.. Also in some other places:

```
lib-imap/imap-capability-list.c:                        /*  now check if IMAP_CAP_VISIBILITY_REQUIRE_ALL is present */
lib-imap/imap-capability-list.c:                                /*  did we match all? */
lib-imap/imap-capability-list.c:                        /*  no IMAP_CAP_VISIBILITY_REQUIRE_ALL - did we match any flags? */
lib-imap/imap-capability-list.c:                        /*  should we add this entry? */
lib-imap/imap-capability-list.h:/*  the initial size of the capability list
lib-imap/imap-capability-list.h:/*  The amount to grow the capability list
lib-imap/imap-capability-list.h:/*  This structure describes a capability word that appears
```

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,85 @@
...
+	IMAP_CAP_VISIBILITY_NO_LOGIN        = BIT(6), /* show this cap if login is disabled */
+	/* NOTE: connections from localhost are 'secure' regardless of SSL/TLS! */
+	IMAP_CAP_VISIBILITY_SECURE          = BIT(7), /* show this cap if we are secured */
+	IMAP_CAP_VISIBILITY_INSECURE        = BIT(8), /* show this cap if we aren't secured */
+
+	/* IMAP_CAP_VISIBILITY_REQUIRE_ALL indicates that the rest
+	   of the conditions must be met *exactly* to show this
+	   capability, otherwise if *at least* 1 condition is met
+	   the capability will be shown */
+	IMAP_CAP_VISIBILITY_REQUIRE_ALL     = BIT(31),
```
sirainen: Standard enums are signed ints. BIT(31) makes it unsigned, which isn't very standard anymore. I suppose should be BIT(30) instead. Also need to change IMAP_CAP_VISIBILITY_ALWAYS to 0x3FFFFFFF then.

Also, this would be better named with a "FLAG" in its name: IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,85 @@
...
+
+	IMAP_CAP_VISIBILITY_ALWAYS          = 0x7FFFFFFF /* always show this cap */
+};
+
+/*  the initial size of the capability list
+	it gets automatically increased but this
+	is the size that we start at */
+#define INITIAL_CAP_LIST_SIZE 5
```
sirainen: This is not used anywhere. So this is the one I suggested to be 64 earlier. But it's not really needed to be a macro, since it's clear enough anyway. -> remove

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,85 @@
...
+/*  The amount to grow the capability list
+	each time it fills up */
+#define CAP_LIST_GROW_AMOUNT  1
```
sirainen: Unused -> remove

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,85 @@
...
+	   capability, otherwise if *at least* 1 condition is met
+	   the capability will be shown */
+	IMAP_CAP_VISIBILITY_REQUIRE_ALL     = BIT(31),
+
+	IMAP_CAP_VISIBILITY_ALWAYS          = 0x7FFFFFFF /* always show this cap */
+};
+
+/*  the initial size of the capability list
+	it gets automatically increased but this
+	is the size that we start at */
```
sirainen: Indentation inside the comments are a bit weird.

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,85 @@
...
+/*  The amount to grow the capability list
+	each time it fills up */
+#define CAP_LIST_GROW_AMOUNT  1
+
+/*  This structure describes a capability word that appears
+	in the imap and imap-login capability string. The flags
+	indicate when and under which conditions the capability
+	word should be displayed */
+struct imap_capability {
+	char *name;
```
sirainen: const char *name

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,153 @@
...
+	capability_list->pool = pool_alloconly_create(MEMPOOL_GROWING"Capability List", 2048);
+	/* initialize the capability list capability array */
+	p_array_init(&(capability_list->imap_capabilities),
+			capability_list->pool,
+			0);
+
+	/* make sure our initial entries are appended */
+	if (initial_entries != NULL) {
+		imap_capability_list_append_string(capability_list, initial_entries);
+	}
```
sirainen: Coding style consistency: No { and } when there's only a single line in the block.

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,153 @@
...
+			/* strip off the REQUIRE_ALL and any flags that don't match */
+			tmp_flags &= (flags & ~IMAP_CAP_VISIBILITY_REQUIRE_ALL);
+
+			bool add = FALSE;
+			/*  now check if IMAP_CAP_VISIBILITY_REQUIRE_ALL is present */
+			if ((capp->flags & IMAP_CAP_VISIBILITY_REQUIRE_ALL) != 0) {
+				/*  did we match all? */
+				if ((capp->flags & ~IMAP_CAP_VISIBILITY_REQUIRE_ALL) == tmp_flags) {
+					add = TRUE;
+				}
```
sirainen: Coding style consistency: No { and } when there's only a single line in the block.

src/imap/imap-client.c
```diff
@@ -552,8 +558,36 @@ void client_add_capability(struct client *client, const char *capability)
...
+	imap_capability_list_add(client->capability_list,
+			capability, IMAP_CAP_VISIBILITY_ALWAYS);
+}
+
+const char *client_get_capability(struct client *client)
+{
+	string_t *cap_str = t_str_new(256);
+
+	/* imap is postauth by definition */
+	uint32_t flags = IMAP_CAP_VISIBILITY_POSTAUTH;
```
sirainen: `enum imap_capability_visibility visibility = ...`

src/imap-login/imap-login-client.c
```diff
@@ -93,31 +94,31 @@ static const char *get_capability(struct client *client)
...
-	else if (*imap_client->set->imap_capability != '+') {
-		explicit_capability = TRUE;
-		str_append(cap_str, imap_client->set->imap_capability);
-	} else {
-		str_append(cap_str, CAPABILITY_BANNER_STRING);
-		str_append_c(cap_str, ' ');
-		str_append(cap_str, imap_client->set->imap_capability + 1);
-	}
+	/* imap-login is preauth by definition */
+	uint32_t flags = IMAP_CAP_VISIBILITY_PREAUTH;
```
sirainen: `enum imap_capability_visibility visibility = ...`

src/lib-imap/imap-capability-list.h
```diff
@@ -0,0 +1,85 @@
...
+#include "array.h"
+
+/* bitflags to represent different conditions of
+   the current imap session */
+enum imap_capability_visibility {
+	IMAP_CAP_VISIBILITY_NEVER           = 0,      /* never show this capability */
+	IMAP_CAP_VISIBILITY_PREAUTH         = BIT(0), /* show this cap before login */
+	IMAP_CAP_VISIBILITY_POSTAUTH        = BIT(1), /* show this cap after login */
+	IMAP_CAP_VISIBILITY_TLS_ACTIVE      = BIT(2), /* show this cap if tls is active */
+	IMAP_CAP_VISIBILITY_TLS_INACTIVE    = BIT(3), /* show this cap if tls is inactive */
```
sirainen: These are now confusing. Are these set if SSL/TLS is enabled? Or are these set only when STARTTLS is used? Now imap process uses the former way and imap-login uses the latter way. The comments should be at least updated, and if it means just STARTTLS then rename the capabilities to IMAP_CAP_VISIBILITY_STARTTLS_*

src/imap-login/imap-login-client.c
```diff
@@ -389,6 +391,52 @@ static void imap_client_create(struct client *client, void **other_sets)
...
+				CAPABILITY_BANNER_STRING);
+		/* add everything after the plus to our cap list */
+		imap_capability_list_append_string(imap_client->capability_list,
+				imap_client->set->imap_capability + 1);
+	}
+
+	/* add STARTTLS to cap list if it is needed */
+	if (strcmp(client->ssl_set->ssl, "no") != 0) {
+		imap_capability_list_add(imap_client->capability_list,
+				"STARTTLS", IMAP_CAP_VISIBILITY_REQUIRE_ALL |
```
sirainen: I think this now returns STARTTLS when connecting to imaps port. Also should likely use `client_is_tls_enabled(client)` instead of `strcmp(client->ssl_set->ssl, "no") != 0`

src/imap-login/imap-login-client.c
```diff
@@ -389,6 +391,52 @@ static void imap_client_create(struct client *client, void **other_sets)
...
+							IMAP_CAP_VISIBILITY_PREAUTH |
+							IMAP_CAP_VISIBILITY_INSECURE);
+	}
+
+	/* add LOGINDISABLED to cap list */
+	imap_capability_list_add(imap_client->capability_list,
+			"LOGINDISABLED", IMAP_CAP_VISIBILITY_REQUIRE_ALL |
+							 IMAP_CAP_VISIBILITY_PREAUTH |
+							 IMAP_CAP_VISIBILITY_NO_LOGIN);
```
sirainen: Coding style consistency: Indentation here and elsewhere would be better as:
```
	imap_capability_list_add(imap_client->capability_list, "LOGINDISABLED",
		IMAP_CAP_VISIBILITY_REQUIRE_ALL |
		IMAP_CAP_VISIBILITY_PREAUTH |
		IMAP_CAP_VISIBILITY_NO_LOGIN);
```

(or if needed maybe put the strings to the second line with similar indentation)

src/lib-imap/imap-capability-list.c
```diff
@@ -37,13 +37,14 @@ void imap_capability_list_unref(struct imap_capability_list **capability_list)
 {
 	i_assert((*capability_list)->refcount > 0);
 
-	*capability_list = NULL;
-
-	if (--(*capability_list)->refcount > 0)
+	if (--(*capability_list)->refcount > 0) {
+		*capability_list = NULL;
```
sirainen: These lib-imap changes seem to be in a wrong commit (imap-login commit).

src/lib-imap/imap-capability-list.c
```diff
@@ -0,0 +1,153 @@
...
+void imap_capability_list_append_string(struct imap_capability_list *capability_list,
+		const char *name_list)
+{
+	T_BEGIN {
+		/* split the name list by space */
+		const char **strings = t_strsplit_spaces(name_list, " ");
+		/* iterate each name and append it to list as ALWAYS */
+		while (*strings != NULL) {
+			imap_capability_list_add(capability_list, *strings, 
+					IMAP_CAP_VISIBILITY_ALWAYS);
```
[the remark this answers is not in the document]
shaun-a-johnson: Hmm.. I'm not sure I understand what you mean by "a single tab at the beginning of imap_capability_list_add()"

You mean adding another tab before "imap_capability_list_add()"? Like this:
`
				imap_capability_list_add(capability_list, *strings, 
					IMAP_CAP_VISIBILITY_ALWAYS);
`

Such that there is 4 tabs before imap_capability_list_add()?

src/lib-imap/imap-capability-list.h
shaun-a-johnson: Ah yes this was originally why I used the uint32_t in the lookup function (because of the sign bit), I will change it to use bit 30 instead.

I assume you also want to see "FLAG" in the other capability flags, I will add that as well.

src/imap-login/imap-login-client.c
shaun-a-johnson: > I think this now returns STARTTLS when connecting to imaps port.
Is there any changes I need to do regarding this?

Other than that, I have updated the strcmp to client_is_tls_enabled().

src/imap-login/imap-login-client.c
shaun-a-johnson: I may have misunderstood previous directions, are the parameters being passed to function calls supposed to line up with the "("? 

Or is that only for function definitions/declarations?

I only ask because the example code you give above doesn't seem to line up the capability flag parameters with the "(", but I'm not sure if that's just github rendering the tabs a certain way or not.

src/lib-imap/imap-capability-list.h
sirainen: No, to me it seems like REQUIRE_ALL is special in a way from the other enums here? That's why I was thinking about the FLAG in it.

src/lib-imap/imap-capability-list.c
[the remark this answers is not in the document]
sirainen: I mean usually we do:
```
			imap_capability_list_add(capability_list, *strings, 
						 IMAP_CAP_VISIBILITY_ALWAYS);
```

But if that would bring the line length (a lot) over 80 chars, we do also:

```
			imap_capability_list_add(capability_list, *strings, 
				IMAP_CAP_VISIBILITY_ALWAYS);
```

src/imap-login/imap-login-client.c
sirainen: Actually it seems to be a bit wrong. Anyway my main concern was that the original code was:

```
	if (client_is_tls_enabled(client) && !client->tls)
		str_append(cap_str, " STARTTLS");
```

But your code isn't using `client->tls` anywhere. It seems you're actually using `client->starttls`. So I think now if you connect to localhost the CAPABILITY list won't show STARTTLS?

src/imap-login/imap-login-client.c
sirainen: Maybe the earlier comment I just sent made it clearer? It could line up to '(' as long as the line lengths don't go (much) over 80 chars. I'm not sure if they did here or not. Both are acceptable anyway :)

src/lib-imap/imap-capability-list.h
shaun-a-johnson: The flag was originally accompanied by two other flags: 

     IMAP_CAP_VISIBILITY_SSL_ACTIVE
     IMAP_CAP_VISIBILITY_SSL_INACTIVE

However it was raised in a previous review that the differentiation between SSL and TLS was not appropriate, and to remove the SSL flag. 

So given that decision, it would make sense for the remaining flag `IMAP_CAP_VISIBILITY_TLS_ACTIVE` to refer to either one: IMAPS or STARTTLS.

This flag allows for a session to be differentiated between:
 - Secured for any reason (IMAP_CAP_VISIBILITY_SECURE)
 - Secured specifically with TLS/SSL (IMAP_CAP_VISIBILITY_TLS_ACTIVE)

Should the flag still be renamed if it does not correspond directly to STARTTLS?

sirainen: Then it's okay, but the comments in them could say "SSL/TLS" instead of just "tls". And also in imap-login the code is now:

```
	if (client->starttls)
		flags |= IMAP_CAP_VISIBILITY_TLS_ACTIVE;
```

So it should be instead:

```
	if (client->tls)
		flags |= IMAP_CAP_VISIBILITY_TLS_ACTIVE;
```

Although now I see there's nowadays something called client->ssl_secured .. I'm not really sure why we have a difference between that and client->tls. Have to look into that.

src/lib-imap/imap-capability-list.h
shaun-a-johnson: Understood, I will add FLAG to the REQUIRE_ALL capability, and leave the rest.

src/lib-imap/imap-capability-list.h
shaun-a-johnson: If my memory serves from when I was investigating those variables (forgive me if I am mistaken, I am no expert):

 - client->starttls is whether imap-login is using STARTTLS
 - client->tls is whether imap-login is using any SSL/TLS/STARTTLS
 - client->ssl_secured is the flag that is passed along to the post-login binary (imap) to indicate whether SSL is active or not (in any form). I believe the post-login (imap) binary has no way of determining whether STARTTLS was used or not -- only whether SSL is effectively active.

sirainen: client->ssl_secured also exists in imap-login. That seems to be used mainly by haproxy support .. I think we'll need to clean up a bunch of this code. Anyway I think client->tls is what you should be using here.

src/imap-login/imap-login-client.c
shaun-a-johnson: Hmmm... I think I understand what you're saying...

The STARTTLS capability is being registered with:

                   imap_capability_list_add(imap_client->capability_list, "STARTTLS",
                                            IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL |
                                            IMAP_CAP_VISIBILITY_PREAUTH |
                                            IMAP_CAP_VISIBILITY_INSECURE);

This means that STARTTLS will be shown if the connection is INSECURE, and we are at PREAUTH.

We still want to display STARTTLS over secured connections like localhost, just not when TLS/SSL is already established in some form.

So would it work if we changed the registration to:

                   imap_capability_list_add(imap_client->capability_list, "STARTTLS",
                                            IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL |
                                            IMAP_CAP_VISIBILITY_PREAUTH |
                                            IMAP_CAP_VISIBILITY_TLS_INACTIVE);

That way in the get_capability function we will have:

        /* is tls/ssl enabled? */
        if (client->tls)
                flags |= IMAP_CAP_VISIBILITY_TLS_ACTIVE;
        else
                flags |= IMAP_CAP_VISIBILITY_TLS_INACTIVE;

The above will effectively perform the `!client->tls` check from the original code, and then the other half of the check is performed at the time of registering the STARTTLS keyword:

        if (client_is_tls_enabled(client)) { 
                imap_capability_list_add(imap_client->capability_list, "STARTTLS",
                                         IMAP_CAP_VISIBILITY_FLAG_REQUIRE_ALL |
                                         IMAP_CAP_VISIBILITY_PREAUTH |
                                         IMAP_CAP_VISIBILITY_TLS_INACTIVE);
        }

So as long as TLS is enabled we will add STARTTLS to the capability list under the condition that it is only ever shown at: preauth when TLS/SSL is not active whatsoever.

Is this an accurate solution to the problem you are describing?

shaun-a-johnson:
Thanks for taking the time to review this Timo, I have pushed another branch which I believe should solve all of the points you have listed.

There is one outstanding question above regarding the STARTTLS capability, I believe I have solved the issue you described but I will leave that issue open for further discussion on that topic.

I will now look to generating automated tests for this, I will get post an update if I run into any issues.

shaun-a-johnson:
[capabilities_tests.tar.gz](https://github.com/dovecot/core/files/3412818/capabilities_tests.tar.gz)

I've attached to this post a tarball containing a very simple bash script and a series of imap test files that can be used in conjunction with the imaptest program.

The bash script will modify the dovecot configuration to cause changes to the capability, reload dovecot, and then run the respective imaptest unit file for the difference in capability.

I may have missed some, but what I gathered was these are all the settings I can manipulate to trigger capability banner changes:

        - ssl = no

                This causes STARTTLS to no longer appear at pre-login

        - imap_literal_minus = yes
        
                This results in LITERAL- instead of LITERAL+

        - mailbox_list_index = no

                This results in the NOTIFY capability disappearing

        - imap_urlauth_host = asdf
          mail_attribute_dict = asdf

                These two cause URLAUTH and URLAUTH=BINARY to appear

        - imap_metadata = 1
        
                This causes METADATA to appear
         
        - *commented out all special_use mailboxes in 15-mailboxes.conf*

                This causes SPECIAL-USE to disappear

        - auth_mechanisms = login

                This causes LOGINDISABLED to appear when connecting over localhost.

The result of running the test script against the upstream master should give the same output as running the test script against this patch.

The expected output is:
```
1 test groups: 0 failed, 0 skipped due to missing capabilities
base protocol: 0/4 individual commands failed
extensions: 0/0 individual commands failed
1 test groups: 0 failed, 0 skipped due to missing capabilities
base protocol: 0/4 individual commands failed
extensions: 0/0 individual commands failed
1 test groups: 0 failed, 0 skipped due to missing capabilities
base protocol: 0/4 individual commands failed
extensions: 0/0 individual commands failed
1 test groups: 0 failed, 0 skipped due to missing capabilities
base protocol: 0/4 individual commands failed
extensions: 0/0 individual commands failed
1 test groups: 0 failed, 0 skipped due to missing capabilities
base protocol: 0/4 individual commands failed
extensions: 0/0 individual commands failed
1 test groups: 0 failed, 0 skipped due to missing capabilities
base protocol: 0/4 individual commands failed
extensions: 0/0 individual commands failed
1 test groups: 0 failed, 0 skipped due to missing capabilities
base protocol: 0/4 individual commands failed
extensions: 0/0 individual commands failed
1 test groups: 0 failed, 0 skipped due to missing capabilities
base protocol: 0/2 individual commands failed
extensions: 0/0 individual commands failed

```

Hopefully this should help to expedite the testing process for this patch, if there was anything I've missed or handled incorrectly here then please let me know.

src/imap-login/imap-login-client.c
shaun-a-johnson: Just wanted to follow up - haven't seen any response on the proposed solution - can you please let me know if what has been proposed is an accurate solution?

Thanks!

sirainen:
Waiting for extra time / clear use case..

kmcgrail:
I like the idea of this.  Seems necessary to move dovecot to more of a plugin-based concept that if capabilities are added/removed by a plugin, that has to be surfaced in the plugin chain.

nerdlem:
Just wanted to add a comment in support of this feature. Plugins that can cause the capabilities they provide dynamically seems like a great idea.

cmouse:
Please reopen against main.

PR #87, opened 2018-06-26
Added linux-only fix to nfs_flush_chown_uid

shaun-a-johnson:
Observed with Netapp NFS mounted filesystems that the 'chown' call that is made in nfs-workarounds.c function nfs_flush_chown_uid results in file mode bits S_ISUID and S_ISGID being dropped.

This fix changes nfs_flush_chown_uid on Linux only to use stat + chmod to flush nfs attribute caches instead of the original chown call that could cause drop of S_ISUID/S_ISGID bits.

Also added backwards compatibility for non-linux to use the original chown method.

This branch is split into two commits where:
 - The first commit simply splits the code in preparation for the second commit, there is no changes to behaviour in the first commit
 - The second commit changes the linux-only chown call that was split out in the first commit to a chmod call

Note this is a re-opened pull request of a pull request that I closed because I accidentally performed the changes on the wrong branch:
https://github.com/dovecot/core/pull/79

### DIFF
--- a/src/lib/nfs-workarounds.c
+++ b/src/lib/nfs-workarounds.c
@@ -178,8 +178,10 @@ static void nfs_flush_chown_uid(const char *path)
 		i_error("nfs_flush_chown_uid: stat(%s) failed: %m", path);
 		return;
 	}
-	if (chown(path, uid, (gid_t)-1) < 0) {
-		if (errno == ESTALE || errno == EPERM || errno == ENOENT) {
+	/* we use chmod for this operation since chown has been seen to drop S_UID
+	   and S_GID bits from directory inodes in certain conditions */
+	if (chmod(path, st.st_mode & 07777) < 0) {
+		if (errno == EPERM) {
 			/* attr cache is flushed */
 			return;
 		}
@@ -187,7 +189,8 @@ static void nfs_flush_chown_uid(const char *path)
 			nfs_flush_file_handle_cache_parent_dir(path);
 			return;
 		}
-		i_error("nfs_flush_chown_uid: chown(%s) failed: %m", path);
+		i_error("nfs_flush_chown_uid: chmod(%s, %04o) failed: %m",
+				path, st.st_mode & 07777);
 	}
 #endif
 }
--- a/src/lib/nfs-workarounds.c
+++ b/src/lib/nfs-workarounds.c
@@ -150,6 +150,17 @@ static void nfs_flush_chown_uid(const char *path)
 
 #ifdef ATTRCACHE_FLUSH_CHOWN_UID_1
 	uid = (uid_t)-1;
+	if (chown(path, uid, (gid_t)-1) < 0) {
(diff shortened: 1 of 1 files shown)


src/lib/nfs-workarounds.c
```diff
@@ -146,16 +146,23 @@ int nfs_safe_link(const char *oldpath, const char *newpath, bool links1)
...
-	uid_t uid;
-
 #ifdef ATTRCACHE_FLUSH_CHOWN_UID_1
-	uid = (uid_t)-1;
+    if (chown(path, (uid_t)-1, (gid_t)-1) < 0) {
+        if (errno == ESTALE || errno == EPERM) {
+            /* attr cache is flushed */
+            return;
+        }
+        if (likely(errno == ENOENT)) {
```
cmouse: `} else if (`

src/lib/nfs-workarounds.c
```diff
@@ -167,18 +174,23 @@ static void nfs_flush_chown_uid(const char *path)
 		i_error("nfs_flush_chown_uid: stat(%s) failed: %m", path);
 		return;
 	}
-#endif
-	if (chown(path, uid, (gid_t)-1) < 0) {
```
cmouse: This should be in another commit, not same commit.

src/lib/nfs-workarounds.c
shaun-a-johnson: Thank you for the review, I have adjusted the if-statements inside the function to use `} else if (`

src/lib/nfs-workarounds.c
shaun-a-johnson: Hello and thank you for taking the time to review this pull request, I would like to apply the fixes you have suggested here but I will need some clarification regarding which piece should be separated into another commit.

When making the commit I considered the entire change to be a singular compartmentalized fix which changed the Linux implementation to use stat + chmod and switched the non-Linux implementations to retain the old chown mechanism. 

Am I correct in my assumption that the change for Linux and change for non-Linux should be separated into individual commits?

shaun-a-johnson:
Hello Dovecot maintainers, 

I have pushed an updated version of this branch which attempts to split the change into two separate commits.

In the first commit I have simply split the chown call into two separate chown calls, previously both linux and freebsd/solaris would fall through to the same chown -- the commit introduces an individual chown into the freebsd/solaris ifdef and it moves the previously existing catch-all chown into the linux-only section of the ifdef.

Effectively the first commit does not change any behaviour it simply formats the code to prepare for the second commit.

The second commit replaces the linux-only chown call with a chmod call that will result in an nfs attribute cache flush without stripping away the sticky bits of the file.

shaun-a-johnson:
cmouse - not sure if you saw the last updates made?  Was there anything else I am missing that you were expecting to see changed?

cmouse:
merged externally.

PR #88, opened 2018-07-02
Fix assert-crash in test suite on i686

pghmcfc:
Fix compilation warnings in test-compression.c due to mismatches
between size_t and uoff_t, which then manifests in assert-crashes
running the test suite on i686 systems.


sirainen:
Looks ok. I'll do a couple of changes to commit message before committing:
 * Prefix with "lib-compression: "
 * i686 -> 32bit

pghmcfc:
No problem. I suspected it was a general 32-bit problem but I've only got i686 to test with.

PR #89, opened 2018-07-15
Add iterate support for userdb vpopmail

gutweiler:
This adds user iteration support for vpopmail.

Tested with doveadm user '*' and together with replicator/dsync.

Alias domains of vpopmail will be not returned, because that would result in duplicates.

```
$ /usr/local/dovecot/bin/doveadm user '*'
postmaster@example1.com
test@example.com

$ /usr/local/dovecot/bin/doveadm replicator status '*'
username                    priority fast sync full sync success sync failed
postmaster@example.com      none     00:00:56  00:05:56  00:00:56     -
test@example.com            none     00:01:32  00:05:56  00:01:31     -
```

### DIFF
--- a/src/auth/userdb-vpopmail.c
+++ b/src/auth/userdb-vpopmail.c
@@ -17,6 +17,16 @@ struct vpopmail_userdb_module {
 	const char *quota_template_value;
 };
 
+struct vpopmail_iterate_context {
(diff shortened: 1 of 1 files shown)


cmouse:
vpopmail is no longer supported or maintained by us.

PR #90, opened 2018-09-27
Fix XCLIENT PORT capability usage

antondollmaier:
If Postfix advertises "XCLIENT NAME ADDR PROTO HELO REVERSE_NAME PORT LOGIN", submission currently fails: Postfix is then unable to parse the XCLIENT request from Dovecot Submissiond:

```
< XCLIENT ADDR=185.115.176.12PORT=39074
> 501 5.5.4 Bad ADDR syntax: 185.115.176.12PORT=39074
```

This is my first pull request and I was yet unable to verify the changes - but as line 69 (FORWARD=) also implies the same logic, I'm confident that this patch is usable.

### DIFF
--- a/src/submission-login/submission-proxy.c
+++ b/src/submission-login/submission-proxy.c
@@ -51,19 +51,19 @@ proxy_send_xclient(struct submission_client *client, struct ostream *output)
 	}
 
 	str = t_str_new(128);
-	str_append(str, "XCLIENT ");
+	str_append(str, "XCLIENT");
 	if (str_array_icase_find(client->proxy_xclient, "ADDR")) {
-		str_append(str, "ADDR=");
+		str_append(str, " ADDR=");
 		str_append(str, net_ip2addr(&client->common.ip));
 	}
 	if (str_array_icase_find(client->proxy_xclient, "PORT"))
-		str_printfa(str, "PORT=%u", client->common.remote_port);
+		str_printfa(str, " PORT=%u", client->common.remote_port);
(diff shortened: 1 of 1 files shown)


stephanbosch:
Although you found a valid problem, the fact that your Postfix MTA is getting to see this tells me you're using the submission service wrong (which I've seen more often now). The code you modified is supposed to be only involved in proxying between two dovecots (e.g. director proxy and storage backend), so not between Dovecot and the relay MTA. Configure the relay MTA using the submission_relay_* settings (https://wiki2.dovecot.org/Submission#Relay_MTA) and *NOT* using login proxy settings from passdb (https://wiki2.dovecot.org/PasswordDatabase/ExtraFields/Proxy). I added a notice to https://wiki2.dovecot.org/Submission#Login_Proxy. Please tell me if anything is still unclear.

I will create an internal merge request for your fix and it end up in master somewhat soon.

antondollmaier:
Hi,

> Although you found a valid problem, the fact that your Postfix MTA is getting to see this tells me you're using the submission service wrong (which I've seen more often now). The code you modified is supposed to be only involved in proxying between two dovecots (e.g. director proxy and storage backend), so not between Dovecot and the relay MTA. 

Understood - I have added a submissiond instance on the storage backend and configured the proxy to use this. The error appeared here as well.

> I will create an internal merge request for your fix and it will end up in master somewhat soon.

Thanks!

sirainen:
Merged.

PR #92, opened 2018-10-06
userdb-passwd: set errno=0 before each getpwent()

trofi:
In https://bugs.gentoo.org/667118 Reuben Farrelly
noticed that running
    # doveadm user '*'
causes auth daemon to generate errors like:
    auth-worker(3585): Error: getpwent() failed: Invalid argument

This happens because on successful call getpwent()
now sets errno=EINVAL starting from glibc-2.28.
See https://sourceware.org/PR16004 for details.

The fix is to check 'errno' only when 'getpwent()' fails.

Reported-by: Reuben Farrelly
Bug: https://bugs.gentoo.org/667118
Bug: https://sourceware.org/PR16004
Signed-off-by: Sergei Trofimovich <slyfox@gentoo.org>

### DIFF
--- a/src/auth/userdb-passwd.c
+++ b/src/auth/userdb-passwd.c
@@ -172,8 +172,11 @@ static void passwd_iterate_next(struct userdb_iterate_context *_ctx)
 		return;
 	}
 
-	errno = 0;
-	while ((pw = getpwent()) != NULL) {
+	for (;;) {
+		errno = 0;
+		pw = getpwent();
+		if (pw == NULL)
+			break;
 		if (passwd_iterate_want_pw(pw, set)) {
 			_ctx->callback(pw->pw_name, _ctx->context);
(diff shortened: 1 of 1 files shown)


reubenfarrelly:
This has resolved the error for me - and I've been running this for some weeks now with no unexpected effects.  Can this be tested/merged/pulled in for the next release?

cmouse:
It's bit late for next release, but maybe after that..

cmouse:
I'll check where it's going.

sirainen:
Merged with slightly different code: https://github.com/dovecot/core/commit/414a7db1d15e57433dfc222bd6bf253c83166408

Could you verify that we didn't mess it up somehow? :)

reubenfarrelly:
Looks all good to me - tested just fine.

PR #95, opened 2018-11-26
Aliases

javerous:
**Context**

- Dovecot is configured as a submission + imap service. It interacts with postfix with LMTP for mail box distribution and with submission relay for mail relaying.
- Postfix is configured as a server-to-server mail distribution service. It doesn’t do identification or authentication by itself.
- The whole mail system is used for private e-mailing, with small user base. It use a simple "passwd-file" driver.

**Purpose**

All identification and authentication is done by dovecot : when a mail client connects to imap, or when postfix want to distribute an e-mail to a local mail box.

I find it a bit unfortunate to "break" authentication encapsulation, and having postfix to handle aliasing to dovecot defined users.
So I did this modification in the code to be able to define aliases for users in the passwd-file files.

It works by adding at least one field named "aliases" on extra field, and by setting an alias user list, separated by `,` character.

When dovecot needs to do an authentication lookup, if the user is not found, it will try aliases to find the real user, and use this one as final authentication. So it's possible to login to imap with aliases, and send e-mail to alias to have them distributed in the real single user mail box.

I think it's easier to use than the MailboxAlias plugin, based on symbolic links.

I use it internally since a few days, and it seems to works as expected.

**Example**

An entry from "passwd-file" can look like this :
```
user:(password):_unknown:_unknown:(gecos):(home):(shell):aliases=alias1,alias2,alias3
```

**Weakness**

- It's perhaps more a personal grade patch than a PR grade change. The intend is mainly to push a bootstrap for a wide alias feature in dovecot than really include the current state of this patch.
- It always (re)set the authentication request username, even if it didn't change because of aliasing.
- The "aliases" extra-field is not prefixed with "userdb_".

### DIFF
--- a/src/auth/db-passwd-file.c
+++ b/src/auth/db-passwd-file.c
@@ -40,9 +40,17 @@ passwd_file_add(struct passwd_file *pw, const char *username,
 			pw->path, username);
 		return;
 	}
+	
+	if (hash_table_lookup(pw->aliases, username) != NULL) {
+		i_error("passwd-file %s: User %s already defined as an alias",
+				pw->path, username);
+		return;
+	}
 
 	pu = p_new(pw->pool, struct passwd_user, 1);
 	user = p_strdup(pw->pool, username);
+
+	pu->username = user;
 
 	len = pass == NULL ? 0 : strlen(pass);
 	if (len > 4 && pass[0] != '{' && pass[0] != '$' &&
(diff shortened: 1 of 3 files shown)


cmouse:
Hi!

Thank you for the pull request, however, I believe this can be achieved with a simple config change instead:

```
passdb {
   driver = passwd-file
   args = /path/to/alias.passwd
}
... rest of authentication
```

Then, into alias.passwd, you put 
```
alias:::::::user=actual_username noauthenticate
```

javerous:
Hi @cmouse. 

Isn't going to fail with validation done on parsing (`passwd_file_add`) ? uid, git, password, etc.

I will make a try, but it's perhaps still a possibility.

Anyway, I think I will keep my patch eventually for my personal usage (it seems a bit more “clean” to me), and I will perfectly understand if you reject this PR, or even if you don’t plan to add a dedicated alias mechanism ;)

javerous:
So, I tested. Perhaps I misconfigured, but when I send an e-mail to `alias@domain.tld`, it ends in the `alias` mailbox, and not in `actual_username` mailbox.

With the proposed patch, e-mails ends up in the right mailbox (the "real" user), as if we were using postfix aliasing.

cmouse:
Those fields are required to be eventually provided, not mandatory. The same happens with this, as if you were using postfix aliasing, and it even looks the same. Although I'd personally do aliasing in postfix.

cmouse:
if you did the alias passdb correctly, it will distribute the e-mails to correct box.

It is usually much simpler to do the alias resolution already in MTA.

sirainen:
Since this is LMTP, the aliases should more importantly be in the userdb. So I guess:

<pre>
userdb {
   driver = passwd-file
   args = /path/to/alias.passwd
   result_success = continue
}
</pre>

javerous:
@sirainen Yes, I tried that too, without success :s

I use `username_format=%Ln` + `override_fields = mail=xxx`, so perhaps it's related.

Anyway, I feel that there is not a great interest to have a dedicated alias mechanism from @cmouse (I'm okay with that), so I will probably keep that patch for me, and perhaps distribute it on another channel if it happens that someone is interested.

cmouse:
Thank your for your effort, but I'm closing this.

PR #96, opened 2019-01-02
Fix several OpenSSL compilation problems

neheb:


cmouse:
Thank you for your pull rqeuest, we'll look into it.

cmouse:
merged with changes.

PR #98, opened 2019-01-03
config: reproducible output from settings-get.pl

olafhering:
If the same source gets built twice ('build same source on different
hosts at different times') the resulting files may differ.
Fix this by sorting the hash keys before usage.

Signed-off-by: Olaf Hering <olaf@aepfle.de>


cmouse:
merged externally, thank you for your effort.

PR #99, opened 2019-02-13
Add support for updating dictionary entries stored in Postgresql

geirp:
ON CONFLICT (...) DO UPDATE SET ... added to SQL-statement. 

The pgsql trigger mentioned on https://wiki2.dovecot.org/Quota/Dict is no longer needed.

### DIFF
--- a/src/lib-dict-backend/dict-sql.c
+++ b/src/lib-dict-backend/dict-sql.c
@@ -105,8 +105,9 @@ sql_dict_init(struct dict *driver, const char *uri,
 	i_zero(&sql_set);
 	sql_set.driver = driver->name;
 	sql_set.connect_string = dict->set->connect;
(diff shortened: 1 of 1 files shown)


cmouse:
Since which postgres version is this?

geirp:
This was added in Postgresql 9.5 which was released 2016-01-07.

https://www.postgresql.org/docs/9.5/release-9-5.html

Current release is 11.1.

9.4 is the oldest still supported release and will be supported until February 13, 2020.

https://www.postgresql.org/support/versioning/

slavkoja:
BTW, SQLite has the same syntax supported from 3.24.0 (2018-06-04)

cmouse:
Done as 5329cd3af4a459d12d7710fd1d42fb9a93bd6fe5

PR #102, opened 2019-04-14
fts-solr: added configuration options

johnfawcett:
fts-solr: added configuration options
- batch_size: configure the number of emails sent in single requests to solr, default 1000
- no_soft_commit: avoid sending softCommit to the solr server, default allows softCommit
  If setting it autoSoftCommit on solr should be set to something reasonable (e.g. 60 seconds)
  in file solrconfig.xml
    <autoSoftCommit>
      <maxTime>${solr.autoSoftCommit.maxTime:60000}</maxTime>
    </autoSoftCommit>


cmouse:
I'd change `no_soft_commit` into `soft_commit=yes|no`. Also doveadm_source_ip can also be done with ip routing, e.g. `ip route add dest/32 via gw src ip`, which might be less convenient of course.

johnfawcett:
I'll change the option name as suggested. I'll do it in another pull request, since this one has picked up a later change that is unrelated (the source_ip one).

cmouse:
you can force-push to your branch to drop unwanted commits.

johnfawcett:
updated info about settings

fts-solr: added configuration options

    batch_size: configure the number of emails sent in single requests to solr, default 1000
    soft_commit=yes|no:  send softCommit to the solr server, default yes is backwards compatible behaviour
    If setting to no autoSoftCommit on solr should be set to something reasonable (e.g. 60 seconds)
    in file solrconfig.xml

    ${solr.autoSoftCommit.maxTime:60000}

cmouse:
DOV-3214

cmouse:
merged externally with changes.

PR #103, opened 2019-10-10
Add log_file_mode as a configuration option

r-hartman:
Add log_file_mode as a configuration option to set the mode on a log file when it is first created.


cmouse:
Hi! Thank you for your pull request. Can you provide some explanation/reasoning why this is useful?

r-hartman:
Hi! The reason why I propose this change, is because I came across a situation where a monitoring application runs as an unprivileged user (say user monusr, group mongrp), but needs to be able to read a dovecot log file to watch for certain events or errors.

Setting the dir in which the log file resides to chmod 2750 allowed the log file to be set to mongrp on creation, but it would still require the chmod g+r for the monitoring application to be able to read it.

There are several solutions to this problem but I thought it would be nice to handle this within dovecot's configuration in the same way as the mode can be configured for some (special) files elsewhere in dovecot's configuration.

sirainen:
Once a log file is created, its permissions stay the same. And besides mode someone might want to specify the owner or group. So seems like the easiest fix is to just use chmod/chown after creating the log file. We're trying to avoid adding new settings that aren't really necessary and just make the code more complex.

PR #104, opened 2019-10-12
fix imapc sync_notify

filiphanes:


sirainen:
What does this fix? The sync_notify(0,0) is just a final notification that no more sync_notify() calls are coming for this sync. It's not supposed to make any difference whether the last parameter is 0 or not.

filiphanes:
I understand. As I was looking in imapc code, I couldn't find code which notifies fts plugin about expunges. Is it not implemented yet?

PR #105, opened 2019-11-06
policy server allows empty recipient= in END-OF-MESSAGE state

jnorell:
Fixes a bug in the quota-status which does not return a reply when a message has multiple recipients.

When a message has multiple recipients, a policy server server will receive an empty "recipient=" list in the END-OF-MESSAGE state, which is mishandled in the current quota-status server.  (With a single message recipient, the recipient= is specified just as it is with the RCPT state, and that is handled correctly.)

### DIFF
--- a/src/plugins/quota/quota-status.c
+++ b/src/plugins/quota/quota-status.c
@@ -154,6 +154,7 @@ static int client_input_line(struct connection *conn, const char *line)
 {
 	struct quota_client *client = (struct quota_client *)conn;
 	const char *error;
+	static char *value = "";
 
 	if (*line == '\0') {
 		o_stream_cork(conn->output);
@@ -162,12 +163,24 @@ static int client_input_line(struct connection *conn, const char *line)
 		client_reset(client);
 		return 1;
 	}
+	if (str_begins(line, "protocol_state=")) {
+		value = t_strdup(line + 15);
+		if (strcmp(value, "RCPT") != 0 && strcmp(value, "END-OF-MESSAGE") != 0) {
+			return 1;
+		}
+	}
 	if (client->recipient == NULL &&
 	    str_begins(line, "recipient=")) {
 		if (smtp_address_parse_path(default_pool, line + 10,
 			SMTP_ADDRESS_PARSE_FLAG_ALLOW_LOCALPART |
 			SMTP_ADDRESS_PARSE_FLAG_BRACKETS_OPTIONAL,
 			&client->recipient, &error) < 0) {
+			if (strcmp(value, "END-OF-MESSAGE") == 0) {
+				if (strlen(line) == 10) { // multiple recipients has empty "recipient="
+					client->recipient = NULL;
+					return 1;
+				}
+			}
 			i_error("quota-status: "
 				"Client sent invalid recipient address: %s",
 				error);
--- a/src/plugins/quota/quota-status.c
+++ b/src/plugins/quota/quota-status.c
@@ -175,16 +175,16 @@ static int client_input_line(struct connection *conn, const char *line)
 			SMTP_ADDRESS_PARSE_FLAG_ALLOW_LOCALPART |
 			SMTP_ADDRESS_PARSE_FLAG_BRACKETS_OPTIONAL,
 			&client->recipient, &error) < 0) {
-			if (strcmp(value, "END-OF-MESSAGE") == 0) {
-				if (strlen(line) == 10) { // multiple recipients has empty "recipient="
-					client->recipient = NULL;
-					return 1;
+				if (strcmp(value, "END-OF-MESSAGE") == 0) {
+					if (strlen(line) == 10) { // multiple recipients has empty "recipient="
+						client->recipient = NULL;
+						return 1;
+					}
 				}
-			}
-			i_error("quota-status: "
-				"Client sent invalid recipient address: %s",
-				error);
-			return 0;
+				i_error("quota-status: "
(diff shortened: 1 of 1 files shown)


cmouse:
Thank you for your merge request, we'll look into this!

src/plugins/quota/quota-status.c
```diff
@@ -162,12 +163,24 @@ static int client_input_line(struct connection *conn, const char *line)
...
+			return 1;
+		}
+	}
 	if (client->recipient == NULL &&
 	    str_begins(line, "recipient=")) {
 		if (smtp_address_parse_path(default_pool, line + 10,
 			SMTP_ADDRESS_PARSE_FLAG_ALLOW_LOCALPART |
 			SMTP_ADDRESS_PARSE_FLAG_BRACKETS_OPTIONAL,
 			&client->recipient, &error) < 0) {
+			if (strcmp(value, "END-OF-MESSAGE") == 0) {
```
cmouse: shouldn't END-OF-MESSAGE be considered termination in any case? the strlen below seems superfluous

cmouse: also this should be indented more.

jnorell: > shouldn't END-OF-MESSAGE be considered termination in any case? the strlen below seems superfluous

The strlen is needed, as it is checking the `recipient=` line; the `value` variable (checked for END-OF-MESSAGE on line 178) is the value of the `protocol_state=` line.  Maybe `value` should be renamed to something more self descriptive?

jnorell: > the strlen below seems superfluous

To clarify, the strlen is checking for an empty `recipient=` line, with no recipient address.  The preceding `smtp_address_parse_path()` fails on an empty `recipient=`, but that is valid and always the case in `END-OF-MESSAGE` state when the message has multiple recipients.

jnorell: > shouldn't END-OF-MESSAGE be considered termination in any case?

Note this `END-OF-MESSAGE` is not EOF, signalling end of a stream or such, it is a literal string which is a defined protocol_state for postfix policy server (http://www.postfix.org/SMTPD_POLICY_README.html).

cmouse: ok

stephanbosch: I'm looking into getting this merged, but I have one question:
Why are you calling the quota status policy service from smtpd_end_of_message_restrictions?
In general, using this service for multiple recipients just works for each individual recipient with the smtpd_recipient_restrictions. It only fails when you use the end_of_message one. Why is that useful? When there are multiple recipients, no recipient value is available at that state and there is nothing to lookup from Dovecot. So, what use case are we looking at here?

Anyway, this situation needs to be handled. But if calling it from that acl is not useful I may just as well prohibit - or better yet - ignore it altogether.

jnorell: Hello @stephanbosch,

I originally configured quota-status in the same manner as a quota policy server for another mail system (dbmail) which did utilize information from both RCPT and END-OF-MESSAGE stages.  Since recipients are always available at RCPT, and size is always available at END-OF-MESSAGE, it would correlate the two, so if a message isn't rejected at RCPT it would be checked for each recipient again at END-OF-MESSAGE when the size was known, so that incoming mail did not push the recipient over quota.

I initially thought dovecot's quota-status had a fairly glaring deficiency here, but after reporting this issue I reconsidered dovecot's approach and realize that it does function correctly if used in only the RCPT stage, it simply implements the solution differently.  (Ie. it allows a message delivery to exceed a recipient's quota, then subsequent messages will be rejected.)

That's the "why".  I concur, the error here should be fixed, but if you simply don't use at END-OF-MESSAGE state at all, everything is fine.

litinoveweedle: Hello,
while I understand both sides here, checking quota after  END-OF-MESSAGE stage in postfix smtpd_end_of_data_restrictions has some benefits, as dovecot over qouta and rejection of following messages is a bit guess work implementation. 

Also there are wide spread HOWTO on internet putting dovecot quota to both smtpd_recipient_restrictions and smtpd_end_of_data_restrictions. I only observer this issue after upgrading to debian buster and I was really lucky to be able to pinpoint it to this bug. (only happen as edge case)

Please is there anything blocking to implement quota check to be also allowed in END-OF-MESSAGE stage as proposed? Thank you.

sirainen:
Could you reopen this against main branch?

jnorell:
> Could you reopen this against main branch?

I don't see a way to reopen this, only to create a new PR (which would loose the above comments), am I overlooking something?

sirainen:
Does this work? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

sirainen:
If that doesn't work, just create a new pull request and link this one.

jnorell:
> Does this work? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

Thanks for the link, but no, I can't change the branch of this pull request, I suspect because it is closed.

According to https://github.community/t/re-opening-pull-request/1337 there should be a button to re-open it, but that button is completely absent (not even disabled).

> If that doesn't work, just create a new pull request and link this one.

Will do.

PR #106, opened 2019-11-19
mdbox: support setting save_date

marcan:
This fixes dsync with an mdbox target, which currently resets save_date for all mails (breaking expiry and the like). The dsync code is fine, but the mdbox backend did not support passing through the save_date.

This patch works for me for dsync, but I'm not sure if this implementation is fully correct; there may also be other cases that need changing too (`mdbox_copy()`?). Comments welcome.

### DIFF
--- a/src/lib-storage/index/dbox-multi/mdbox-save.c
+++ b/src/lib-storage/index/dbox-multi/mdbox-save.c
@@ -24,6 +24,7 @@ struct dbox_save_mail {
 	struct dbox_file_append_context *file_append;
 	uint32_t seq;
 	uint32_t append_offset;
+	time_t save_date;
 	bool written_to_disk;
 };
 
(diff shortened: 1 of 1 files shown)


cmouse:
Thanks! We'll look into this.

src/lib-storage/index/dbox-multi/mdbox-save.c
```diff
@@ -25,6 +25,7 @@ struct dbox_save_mail {
 	uint32_t seq;
 	uint32_t append_offset;
 	bool written_to_disk;
+	time_t save_date;
```
sirainen: Could you move this on top of the bool? Practically won't matter now, but generally we keep bools at the end of structs.

marcan: Done!

cmouse:
Internal review

cmouse:
Merged externally.

PR #107, opened 2019-11-22
size_t_signed.m4: use static assertion instead of run test

rofl0r:
a small step towards cross-compile-ability.

### DIFF
--- a/m4/size_t_signed.m4
+++ b/m4/size_t_signed.m4
@@ -3,14 +3,13 @@ dnl * it's more likely vulnerable to buffer overflows. Anyway, C99 specifies
 dnl * that it's unsigned and only some old systems define it as signed.
 AC_DEFUN([DOVECOT_SIZE_T_SIGNED], [
   AC_CACHE_CHECK([whether size_t is signed],i_cv_signed_size_t,[
-    AC_RUN_IFELSE([AC_LANG_SOURCE([[
+    AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
       #include <sys/types.h>
       #include <stdlib.h>
-      int main() {
-        /* return 0 if we're signed */
-        exit((size_t)(int)-1 <= 0 ? 0 : 1);
-      }
+      int arr[(size_t)-1 > 0 ? 1 : -1];
     ]])],[
+      i_cv_signed_size_t=no
+    ],[
       i_cv_signed_size_t=yes
   
       echo
@@ -23,8 +22,6 @@ AC_DEFUN([DOVECOT_SIZE_T_SIGNED], [
         AC_MSG_ERROR([aborting])
       fi
       echo "..ignoring as requested.."
-    ],[
-      i_cv_signed_size_t=no
     ],[])
   ])
   dnl Note: we check size_t rather than ssize_t here, because on OSX 10.2


cmouse:
some more words could be used to explain why this change is needed and what it does?

rofl0r:
oh, certainly; apologies for not doing so.
the current test compiles a binary and runs it, which is impossible with a crosscompiler (you can't run e.g. a mips binary, as produced by a mips cross compier, on x86).

if you take a look at the currently generated shell code in configure you can see that there's a branch that aborts the configure run when crosscompilation is detected.

the proposed patch uses the "static assert trick" which exploits the fact that an array can't have negative size. it the condition tested evaluates to -1 it causes a compile error rather than requiring to run a binary. this and other things are described in my cross-compile focused article https://sabotage-linux.github.io/blog/5/ .

right now dovecot is one of very few packages in sabotage linux that don't support crosscompilation out-of-the-box *at all*. i worked around it by setting several cache variables so the incompatible tests are skipped, but it would be much preferable to have all the tests written in a cross-compile compatible manner.

this PR is basically a testrun to see if dovecot maintainers are open to improvements in this area, with eventually more fixes to follow if accepted.

rofl0r:
force-pushed an update, i didn't notice the logic was inverted.

cmouse:
If you have other patch(es) waiting, could you provide them already? So we could see what sort of things there are and whether it makes sense to combine some of them?

rofl0r:
i don't have other patches at the moment. as mentioned, i didn't want to invest a lot of effort without previously making sure it wouldn't be wasted.

i do though have predefined macros for the following configure tests in my dovecot recipe
to circumvent crosscompile build from breaking.

```
  i_cv_inotify_works=yes \
  lib_cv_va_copy=yes \
  lib_cv___va_copy=no \
  lib_cv_va_val_copy=no \
  i_cv_posix_fallocate_works=no \
  i_cv_signed_size_t=no \
  i_cv_gmtime_max_time_t=31 \
  i_cv_signed_time_t=yes \
  i_cv_mmap_plays_with_write=yes \
  i_cv_c99_vsnprintf=yes \ 
  i_cv_fd_passing=yes \
```
so these would be the candidates to patch in an analogous fashion, if there's the will to tackle these issues.

sirainen:
I think it would be good to change those that can be converted from runtime tests to compile time tests. But it's not possible for all of these. Actually, I'm not sure if it's possible for anything else except this one you just did? Signed time_t perhaps.

rofl0r:
- i_cv_inotify_works - seems to be removed
- lib_cv_va_copy - can be converted into a compile time check. but va_copy is mandated by the C standard anyway.
- lib_cv___va_copy - likewise.
- lib_cv_va_val_copy - likewise.
- i_cv_posix_fallocate_works - checks for ancient glibc bug. this should either be removed completely as it's not dovecot's job to work around broken libc implementations, or converted into an actual runtime check built into the startup code of the binary.
- i_cv_gmtime_max_time_t - likewise. a check for a bug in a particular solaris release. a bug report to opensolaris should be filed, instead of maintaining cruft. if this is considered important, a startup routine should check it when running dovecot.
- i_cv_signed_time_t - seems to be fixed already
- i_cv_mmap_plays_with_write - seems to be about some bug too; if still needed should be converted into a startup check, because not doing so might produce a misbehaving binary, as build host isn't necessarily host this runs on (imagine users updates libc after having built this binary with an older one).
- i_cv_c99_vsnprintf - dunno, C99 is now 20 year old. if still needed, convert into startup check.
- i_cv_fd_passing - "buggy CMSG macros". probably some ancient bug. again, not dovecot's job to workaround it. i could compile you a list of 1000 bugs that were once in some libc using git log, if you want to protect the user from them you basically need to add a check for every single one (in which case configure needs an hour to run), or provide your own bugfree libc implementation together with dovecot.

sirainen:
I guess many of them could be removed, but some comments:

 * i_cv_gmtime_max_time_t - Only the >40 check is a Solaris bug workaround. The rest of the function was needed for everything else as well, because I don't think any libc supported full 64bit time_t for gmtime(). Although this check is less relevant if timegm() is supported. Anyway, I wouldn't remove this.
 * i_cv_mmap_plays_with_write - This didn't work in OpenBSD. I don't know if they've fixed it nowadays or not. Should check that before removing.
 * i_cv_fd_passing - I'd keep this also, because it's sometimes interesting to compile Dovecot in all kinds of weird places and see if it works. Recently this check has been failing with cygwin and the new Windows/Linux subsystem thingy, although maybe that's already fixed there.
   * The BUGGY_CMSG_MACROS part could be removed though. That was buggy in Linux v2.2.

rofl0r:
can we proceed with this? i'd finally like to get rid of my fork of dovecot from my profile.

cmouse:
Hi! Sorry, we have completely missed this. I can move the patch to our internal git, so you don't have to hang to it.

cmouse:
DOV-4182

sirainen:
Thanks, merged.

PR #108, opened 2019-11-25
Hin patch 2.2.36

zaucker:


cmouse:
Hi! 2.2 is no longer developed. 

please make this against master and also add some explanation what this is about.

zaucker:
Sorry, this was a mistake ... didn’t want to submit it to the dovecot repo.

PR #109, opened 2019-11-26
Update INSTALL.md

zaucker:
Add `./autogen.sh` as initial step

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -4,6 +4,7 @@ Build Instructions
 For most people, the usual:
 
 ```
+./autogen.sh
 ./configure
 make
 sudo make install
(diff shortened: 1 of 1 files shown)


cmouse:
That step is only needed if you install ftom git. Tarballs do not need it.

zaucker:
> That step is only needed if you install ftom git. Tarballs do not need it.

Made it more specific. Feel free to dicard/close the PR if you think this is unnecessary.

sirainen:
Merged a somewhat modified version (and looks like I typoed INSTALL.md in commit): https://github.com/dovecot/core/commit/5c778976c0969ce85dd7034401be127be970107a

PR #110, opened 2019-12-05
WIP: Remove redundant getpwnam() in bsdauth.

kousu:
auth_userokay() already checks that the username is valid, but more than that it allows the flexible OpenBSD auth system which, if the admin has enabled it in login.conf,  allows the user to *choose* their authentication style, for example `yubikey` or `skey` or other customized authenticators. Removing this makes dovecot compatible with one-time passwords from skeyinit or oath-toolkit.

(Though, as an aside, the extra check did protect dovecot from [CVE-2019-19521](https://www.openwall.com/lists/oss-security/2019/12/04/5), so that was good)

See https://man.openbsd.org/auth_userokay and https://man.openbsd.org/login.conf.5.

### DIFF
--- a/src/auth/passdb-bsdauth.c
+++ b/src/auth/passdb-bsdauth.c
@@ -18,20 +18,38 @@ bsdauth_verify_plain(struct auth_request *request, const char *password,
 		    verify_plain_callback_t *callback)
 {
 	const char *type;
-	int result;
+	auth_session_t *result;
 
 	e_debug(authdb_event(request), "lookup");
 
 	/* check if the auth is valid */
 	type = t_strdup_printf("auth-%s", request->service);
-	result = auth_userokay(request->user, NULL, t_strdup_noconst(type),
+	result = auth_usercheck(request->user, NULL, t_strdup_noconst(type),
 			       t_strdup_noconst(password));
 
-	if (result == 0) {
+	if (auth_getstate(result) == 0) {
 		auth_request_log_password_mismatch(request, AUTH_SUBSYS_DB);
 		callback(PASSDB_RESULT_PASSWORD_MISMATCH, request);
 		return;
 	}
+
+	/* make sure we're using the username as defined by the database.
+	 * On OpenBSD, auth "styles" will translate a request->user of "name:style"
+	 * to "name" authenticated via "style"; or the default auth style on a system
+	 * might do something unusual to map any input to any username it wants.
+	 * See authenticate(3).
+	 */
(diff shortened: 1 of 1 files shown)


kousu:
This should probably be changed to use [`auth_usercheck`](https://man.openbsd.org/auth_usercheck) followed by [`auth_getpwd()->pw_name`](https://man.openbsd.org/auth_getpwd). That way the auth backend can fully take care of mapping the inputted username:password -> an on-disk dovecot user.

src/auth/passdb-bsdauth.c
```diff
@@ -17,41 +17,20 @@ static void
 bsdauth_verify_plain(struct auth_request *request, const char *password,
 		    verify_plain_callback_t *callback)
 {
-	struct passwd pw;
 	const char *type;
 	int result;
 
-	e_debug(authdb_event(request), "lookup");
```
sirainen: this shouldn't be removed

sirainen:
> This should probably be changed to use
> 
> ```
> auth_session = auth_usercheck(...);
> ...
> auth_request_set_field(request, "user", auth_getpwd(auth_session)->pw_name, NULL);
> ```
> 
> That way the auth backend can fully take care of mapping the inputted username:password -> an on-disk dovecot user.

Yeah, sounds like a good idea to have the pw_name used for the "user".

sirainen:
Is the bsdauth still useful though? Doesn't PAM work as well nowadays? We had been wondering whether to remove bsdauth support entirely because of that.

kousu:
(sorry I lost this! github often drops emails to my domain for some reason. Maybe my greylisting is eating them and they use an aggressive multi-server thing?)

`bsdauth` is still useful on OpenBSD. OpenBSD does [not have PAM](https://man.openbsd.org/?query=pam&apropos=1&sec=0&arch=default&manpath=OpenBSD-current). NetBSD and FreeBSD have it, it's true. In fact, the motivation for my patch is because the customization that PAM offers OpenBSD instead offers natively through bsdauth "styles" (https://man.openbsd.org/login.conf.5#AUTHENTICATION) but dovecot is unable to use them. I would like to be able to use a one-time code to access my webmail on the go, but I can't because I can't communicate my auth style because dovecot has it hard-coded

kousu:
I've fixed it up, but leaving it as WIP until I have the chance to build it on my mailserver and try it out.

sirainen:
We're planning to remove bsdauth in Dovecot v2.4.

PR #113, opened 2020-01-16
Allow checking certificate fingerprints for ssl connections

zaucker:
This is an add-on we developed for one of our customers. As it has no customer-specific code, I wonder if you'd be willing to merge that into master.

Of course I would make any necessary changes.


cmouse:
I am not sure how this would be useful.

zaucker:
> I am not sure how this would be useful.

This adds the fingerprint of client certificates to the list of parameters to be checked at authentication (`checkpassword_setup_env()` in `db-checkpassed.c`).

Our customer uses fingerprints stored in or removed from their LDAP to authorize/deny IMAP access  based on client certificates. A client is only allowed access if he has both a valid none-expired, none-revoked certificate AND the certificate's fingerprint is in the LDAP server.

PR #114, opened 2020-01-30
man: Fix message UID in doveadm sync -1 example

youam:
Quick doc fix. Should be obvious: source message UID is given as 6 above in L97, so it's probably a typo / edit-artifact in L100.


cmouse:
merged externally, commit 5a09059c140dc08cf1c94553ca391d5c211677c5

PR #115, opened 2020-02-08
Use net_gethostbyname() to resolve hosts in redis/memcached dict URI

eirikrye:
This patch enables the use of host names in the redis and memcached dict URI, for example `redis:host=redis.default.svc.cluster.local`.

This is useful in containerized environments such as Kubernetes, where hard coding an ephemeral IP address is not an option.

### DIFF
--- a/src/lib-dict/dict-memcached-ascii.c
+++ b/src/lib-dict/dict-memcached-ascii.c
@@ -379,6 +379,8 @@ memcached_ascii_dict_init(struct dict *driver, const char *uri,
 			  struct dict **dict_r, const char **error_r)
 {
 	struct memcached_ascii_dict *dict;
+	struct ip_addr *ips;
+	unsigned int ips_count;
 	const char *const *args;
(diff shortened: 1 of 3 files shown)


cmouse:
This should use dns-client.

eirikrye:
Possibly, I am not very familiar with the Dovecot code base.

My impression is that `dns-client` should be used for asynchronous DNS lookups. As this code is only executed during the startup of the `auth-worker` process, I don't believe the blocking DNS lookup is a problem. Introducing the additional complexity of `dns-client` seems to be a bit overkill, particularly since the redis client will not be able to handle any work until it is resolved and connected anyway.

cmouse:
The redis worker is used in many other places too, not just in auth-worker. It can also be used to store quota information, last login information and few other things, so it has to work on those cases too.

cmouse:
Also it should probably use connection.c in the first place, and support things like trying all them IP addresses.

eirikrye:
Alright, I'll close this request.

PR #116, opened 2020-03-13
OSS-fuzz integration

DavidKorczynski:
Hi Contributors!

I have started working on the security of Dovecot and in that regard I would like to integrate the project into OSS-fuzz (https://github.com/google/oss-fuzz). This is a free service offered by Google where they perform continuous fuzzing of open-source projects that are heavily-used by the industry and with the only expectation that maintainers of the project will use and patch the bugs found by OSS-Fuzz. Dovecot has been on the "wanted" list as it's a popular project. In this pull-request I have added three fuzz targets as well as the initial integration into OSS-fuzz (I will give a comment with a link following this PR).
If you are interested in this, could you please provide me with an email address that will be receiving the bug reports that my fuzzers find? Following integration into OSS-fuzz I will make sure more fuzz targets gets developed on Dovecot. 

David

### DIFF
--- a/src/lib-http/fuzz-http-url.c
+++ b/src/lib-http/fuzz-http-url.c
@@ -1,0 +1,35 @@
+
+#include <stdint.h>
+
+#include "lib.h"
+#include "net.h"
+#include "http-url.h"
+#include "test-common.h"
+
+
+int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size);
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    T_BEGIN{
+    struct http_url *urlp = NULL;
+    const char *error = NULL;
+
+    // Allocate string we null terminate
+	char *new_data = malloc(size+1);
+    memcpy(new_data, data, size);
+	if (new_data == NULL)
+        return 0;    
+    new_data[size] = '\0';
+    
+	// Fuzz entrypoint
+    pool_t pdata_stack = pool_datastack_create();
+    http_url_parse(new_data, NULL, 0, pdata_stack, &urlp, &error);
+
+    free(new_data);
+    }T_END;
+
+
+    return 0;
+}
+
--- a/src/lib-http/fuzz-http-url.c
+++ b/src/lib-http/fuzz-http-url.c
@@ -16,17 +16,13 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     const char *error = NULL;
 
     // Allocate string we null terminate
-	char *new_data = malloc(size+1);
-    memcpy(new_data, data, size);
-	if (new_data == NULL)
-        return 0;    
-    new_data[size] = '\0';
+    char *new_data = i_strndup(data, size);
     
 	// Fuzz entrypoint
     pool_t pdata_stack = pool_datastack_create();
     http_url_parse(new_data, NULL, 0, pdata_stack, &urlp, &error);
 
-    free(new_data);
+    i_free(new_data);
     }T_END;
 
 
--- a/src/lib-imap/fuzz-imap-url.c
+++ b/src/lib-imap/fuzz-imap-url.c
@@ -1,0 +1,33 @@
+
+#include <stdint.h>
+
+#include "lib.h"
+#include "net.h"
+#include "imap-url.h"
+#include "test-common.h"
+
+
+int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size);
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    T_BEGIN{
+    struct imap_url *urlp;
+    const char *error = NULL;
+
+    // Allocate new string that we can NULL terminate
+	char *new_data = malloc(size+1);
+    memcpy(new_data, data, size);
+	if (new_data == NULL)
+        return 0;    
+    new_data[size] = '\0';
+    
+    // Fuzz entry
+    imap_url_parse(new_data, NULL, IMAP_URL_PARSE_ALLOW_URLAUTH, &urlp, &error);
+
+    free(new_data);
+    }T_END;
+
+    return 0;
+}
+
--- a/src/lib-imap/fuzz-imap-utf7.c
+++ b/src/lib-imap/fuzz-imap-utf7.c
@@ -1,0 +1,38 @@
+#include <stdint.h>
+
+#include "lib.h"
+#include "str.h"
+#include "test-common.h"
+#include "unichar.h"
+#include "imap-utf7.h"
+#include "test-common.h"
+#include "imap-quote.h"
+
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    if (size < 3)
+        return 0;
+
+	T_BEGIN{
+	string_t *dest;
+	dest = t_str_new(size);
+
+    // allocate new data we NULL terminate
+    char *new_data = malloc(size+1);
+    memcpy(new_data, data, size);
+    new_data[size] = '\0';
+
+    // Fuzz entries
+	imap_utf8_to_utf7(new_data, dest);
+	imap_utf7_to_utf8(new_data, dest);
+	imap_append_string_for_humans(dest, (const unsigned char *)new_data, size);
+
+	str_free(&dest);
+    free(new_data);
+	}T_END;
+
+    return 0;
+}
+


cmouse:
Hi! 

We will look into this and see where the mails could be sent.

sirainen:
I'd be very interested in getting good fuzzing into Dovecot. But I feel like there should be some kind of a cleaner framework for it. If I understand correctly, the ideal way to implement fuzzing would be:
 * Have a separate .c file for each function that takes untrusted user input (I'm not sure why you put 3 separate calls into fuzz-imap-utf7.c?)
 * Have a bunch of valid and invalid test input for each such binary. The more the better? Seems like these could become rather huge.

Would all of this be better in a separate git repository with one directory per test (one .c file and a bunch of sample tests)?

I was also first wondering if this could be integrated into Dovecot's current unit tests in some way, but that's probably not going to work.

src/lib-http/fuzz-http-url.c
```diff
@@ -0,0 +1,35 @@
...
+    T_BEGIN{
+    struct http_url *urlp = NULL;
+    const char *error = NULL;
+
+    // Allocate string we null terminate
+	char *new_data = malloc(size+1);
+    memcpy(new_data, data, size);
+	if (new_data == NULL)
+        return 0;    
+    new_data[size] = '\0';
```
sirainen: All this could be done more easily with `char *new_data = i_strndup(data, size); ... i_free(new_data);` or since you have the T_BEGIN/T_END anyway could also be `const char *new_data = t_strndup(data, size);`

DavidKorczynski:
Awesome!
I tried to follow the pattern of how your test files work, e.g. `test-http-url.c` ==> `fuzz-http-url.c`  
Regarding the smarter architecture around it then I am all for it and would be happy to do/help with an ideal integration, although this will take some time.

Direct answers to your questions:
- The reason I have 3 separate calls into fuzz-imap-utf7.c is because it makes sense to "merge" computation together if we target continuous fuzzing. The OSS-fuzz project will allocate some computing resources per fuzz-target, and, therefore, if fuzz-targets are small to medium in size it's more convenient just allocating a single CPU rather than multiple. The utf fuzzer has lower coverage than the url fuzzers, and that's the reason I merged multiple calls together. Some reasoning behind this is given here: https://github.com/google/oss-fuzz/issues/2469#issuecomment-497069346

- We don't really need to have good/bad inputs in combination with excepted results, but rather let the fuzzers do their work. The current fuzzers that I have added already executes a decent amount of functionality even with the fuzzers operating by themselves. If you mean that we should have some seed corpuses, then definitely that would make sense, although I think this will be more relevant when we go for more complex targets within Dovecot. 

I dont think it is necessary to have it in a seperate git. I could certainly do that if you prefer, but I quite like the setup of having them follow the current test files (`test-http-url.c` ==> `fuzz-http-url.c`) I can manage the fuzzing-side of things and ensure that all is build well and run with the OSS-fuzz project.

DavidKorczynski:
Hey guys, did you have a bit more thoughts on this? if you would like to test out the idea and see how it fits with your way of working we could store the fuzzers within the oss-project itself, e.g. https://github.com/google/oss-fuzz/tree/master/projects/binutils, the only thing I would need is an email where we can send bug reports. I would be happy to continue developing fuzzers for the project on more parts of the code as well and, naturally, down the line it would be nice to integrate the fuzzing into the main project itself.

catenacyber:
Hi @DavidKorczynski,
I have been developing in parallel some fuzzing for Dovecot as well, with a first target for the SMTP server. 
By the way, I saw that you developed additional fuzz targets for binutils which I started ;-)

What is the best way to merge our works ?

catenacyber:
By the way, for the compile scheme, I have been doing this patch :
```
diff --git a/configure.ac b/configure.ac
index 0185039feb..213faa5eb0 100644
--- a/configure.ac
+++ b/configure.ac
@@ -727,6 +727,7 @@ AM_CONDITIONAL(BUILD_MYSQL, test "$build_mysql" = "yes")
 AM_CONDITIONAL(BUILD_SQLITE, test "$build_sqlite" = "yes")
 AM_CONDITIONAL(BUILD_CASSANDRA, test "$build_cassandra" = "yes")
 AM_CONDITIONAL(SQL_PLUGINS, test "$want_sql" = "plugin")
+AM_CONDITIONAL([BUILD_FUZZTARGETS], [test "x$LIB_FUZZING_ENGINE" != "x"])
 
 dnl **
 dnl ** Plugins
diff --git a/src/lib-smtp/Makefile.am b/src/lib-smtp/Makefile.am
index b03761df8b..2d7710394c 100644
--- a/src/lib-smtp/Makefile.am
+++ b/src/lib-smtp/Makefile.am
@@ -83,7 +83,14 @@ test_programs = \
 
 test_nocheck_programs =
 
-noinst_PROGRAMS = $(test_programs) $(test_nocheck_programs)
+fuzz_programs =
+
+if BUILD_FUZZTARGETS
+fuzz_programs += \
+       fuzz-smtp-server
+endif
+
+noinst_PROGRAMS = $(fuzz_programs) $(test_programs) $(test_nocheck_programs)
 
 EXTRA_DIST = \
        test-bin/sendmail-exit-1.sh \
@@ -161,6 +168,11 @@ test_smtp_server_errors_LDFLAGS = -export-dynamic
 test_smtp_server_errors_LDADD = $(test_libs)
 test_smtp_server_errors_DEPENDENCIES = $(test_deps)
 
+fuzz_smtp_server_SOURCES = fuzz-smtp-server.c
+fuzz_smtp_server_LDFLAGS = -export-dynamic $(LIB_FUZZING_ENGINE)
+fuzz_smtp_server_LDADD = $(test_libs)
+fuzz_smtp_server_DEPENDENCIES = $(test_deps)
```

That builds the fuzz programs if environment variable `LIB_FUZZING_ENGINE` is set (as in oss-fuzz)
But then, I am struggling to get clang++ as a linker for fuzz_smtp_server

sirainen:
Here's my initial thoughts on how the framework might work: dovecot/core#117

But I'm having some trouble getting -fsanitize=fuzzer to work with my self-compiled clang, so I couldn't test whether it actually works.

Anyway what do you think of it in general? The idea for macros is to maybe make it possible to run other fuzzers, although I don't know if any of them can be run this same way.

sirainen:
> * We don't really need to have good/bad inputs in combination with excepted results, but rather let the fuzzers do their work. The current fuzzers that I have added already executes a decent amount of functionality even with the fuzzers operating by themselves. If you mean that we should have some seed corpuses, then definitely that would make sense, although I think this will be more relevant when we go for more complex targets within Dovecot.

Right, seed corpuses is what I meant. libfuzzer seems to want a directory of corpuses, so if we have e.g. fuzz-imap-utf7 binary should we then have fuzz-imap-utf7-corpus directory full of files? That's the main part that makes me wonder whether it's a good idea to keep it in the Dovecot core repository. How big are these directories going to get, and how often are they going to be updated. I guess it could be a possibility to keep just the corpuses in a separate repository, but maybe that complicates things too much.

> * The reason I have 3 separate calls into fuzz-imap-utf7.c is because it makes sense to "merge" computation together if we target continuous fuzzing. The OSS-fuzz project will allocate some computing resources per fuzz-target, and, therefore, if fuzz-targets are small to medium in size it's more convenient just allocating a single CPU rather than multiple. The utf fuzzer has lower coverage than the url fuzzers, and that's the reason I merged multiple calls together. Some reasoning behind this is given here: [google/oss-fuzz#2469 (comment)](https://github.com/google/oss-fuzz/issues/2469#issuecomment-497069346)

Okay .. so combining utf7/utf8 could make sense since they could have similar corpuses, but combining those with imap-quote code doesn't make much sense to me anymore. Maybe the lib-imap functions are already too low level and the fuzzing should be done at the level of "imap" binary (e.g. similar to how "cat imap-input | ./imap" already works)? Although even if they're too low-level for oss-fuzz, it could still be nice for running clang-fuzzer manually.

catenacyber:
> Right, seed corpuses is what I meant. libfuzzer seems to want a directory of corpuses, so if we have e.g. fuzz-imap-utf7 binary should we then have fuzz-imap-utf7-corpus directory full of files? That's the main part that makes me wonder whether it's a good idea to keep it in the Dovecot core repository. How big are these directories going to get, and how often are they going to be updated. I guess it could be a possibility to keep just the corpuses in a separate repository, but maybe that complicates things too much.

That is a good question.

You do not _need_ a seed corpus, but it is always better to have one.

Oss-fuzz creates and maintains its own corpus for each fuzz target.

It becomes publicly available after 90 days (before that, you need to get authentified to get it)

So my rule of thumb would be :
- if you already have something ready, use it (such as a text file with a list of commands to be parsed)
- if you do not have something ready, start with the empty corpus, and optimize later

> Maybe the lib-imap functions are already too low level and the fuzzing should be done at the level of "imap" binary

When you get integrated into oss-fuzz, you can have some coverage reports (and remove a fuzz target if it is already covered by a bigger one for instance)

catenacyber:
> But I'm having some trouble getting -fsanitize=fuzzer to work with my self-compiled clang, so I couldn't test whether it actually works.

Maybe it is because it should be sometimes `-fsanitize=fuzzer-no-link`

DavidKorczynski:
> Right, seed corpuses is what I meant. libfuzzer seems to want a directory of corpuses, so if we have e.g. fuzz-imap-utf7 binary should we then have fuzz-imap-utf7-corpus directory full of files? That's the main part that makes me wonder whether it's a good idea to keep it in the Dovecot core repository. How big are these directories going to get, and how often are they going to be updated. I guess it could be a possibility to keep just the corpuses in a separate repository, but maybe that complicates things too much.

The fundamental idea behind having a corpus for the fuzzer is to give structure to the input that the fuzzer generates which would otherwise be difficult for the fuzzer to come up with. As such, if there is little structure in the input, it is in the end not necessary. As such, in the utf-X fuzzers I would probably avoid having corpuses since there is not a lot of structure in the data itself. However, when we get deeper into the project, perhaps to some HTTP parsing, e.g. header-parsing, it would be smart to have some corpus sets with example of HTTP header files, but this need not be anything more than 1-3 files in the corpus set. 

To give you an idea of how a repository would look like if we separate the fuzzing to a different repo you can see an example here: https://github.com/guidovranken/django-fuzzers Notice that the corpus directory there is not really how we would approach it when integrating into OSS-Fuzz, since OSS-Fuzz would generate all of the different input files themselves as long as we only give "templates" when the input data i structured. 

> Okay .. so combining utf7/utf8 could make sense since they could have similar corpuses, but combining those with imap-quote code doesn't make much sense to me anymore. Maybe the lib-imap functions are already too low level and the fuzzing should be done at the level of "imap" binary (e.g. similar to how "cat imap-input | ./imap" already works)? Although even if they're too low-level for oss-fuzz, it could still be nice for running clang-fuzzer manually.

At the end of the day we should determine such decisions based on empirical evidence. The closer to the "Input source" each fuzzer will be able reach more code. However, the deeper we go in the code the more focused the fuzzers can be on a certain set of code, which can be beneficial. OSS-Fuzz has limitations in terms of "minimum size" of the fuzzer, and all of the fuzzers I uploaded are well above the limit in terms of covering code. If we later discover the fuzzers do not find bugs then we should make our decisions based on that.

DavidKorczynski:
> Hi @DavidKorczynski,
> I have been developing in parallel some fuzzing for Dovecot as well, with a first target for the SMTP server.
> By the way, I saw that you developed additional fuzz targets for binutils which I started ;-)
> 
> What is the best way to merge our works ?

Hi :)! Don't think there should be too big of a hassle in merging the code - we should do an initial integration of the fuzzers I uploaded and then just stack on top of it from there.

catenacyber:
> Hi :)! Don't think there should be too big of a hassle in merging the code - we should do an initial integration of the fuzzers I uploaded and then just stack on top of it from there.

Indeed, that should not be too big a hassle.

I think the first thing to do is to get the compilation scheme (ie #117 and `DC_DOVECOT_FUZZER` stuff) ok for everyone

catenacyber:
My best shot to use clang++ for linker for fuzz targets is this heavy override in Makefile.am :
```
fuzz_smtp_server_LINK = $(LIBTOOL) $(AM_V_lt) --tag=CXX \
    $(AM_LIBTOOLFLAGS) $(LIBTOOLFLAGS) --mode=link $(CXXLD) \
    $(AM_CXXFLAGS) $(CXXFLAGS) $(fuzz_smtp_server_LDFLAGS) \
    $(LDFLAGS) -o $@
```

sirainen:
What's the benefit in using clang++ for linking?

DavidKorczynski:
I am not 100% sure but I think it's because of how the Docker images in OSS-Fuzz are created. The runtime libraries used by the fuzzers are compiled with clang, and in particular including the flag `-stdlib=libc++` so we need to link to the same libraries. Forcing linking via clang++ has been an annoyance in the past, though,  and an example of where the maintainters of OSS-Fuzz highlights we should make sure to use `clang++` for linking is written here: https://github.com/google/oss-fuzz/issues/2826

sirainen:
I pushed to https://github.com/dovecot/core/pull/117 some kludging to make it link fuzz-targets with CXXLINK. Seems to work against my test c++ library but I didn't try against real oss-fuzz.

DavidKorczynski:
If you confirm that it works on yours with the flags we described here https://github.com/dovecot/core/pull/117#discussion_r397123607 then you can merge all of it in and I will fix the PR request with OSS-Fuzz https://github.com/google/oss-fuzz/pull/3502 such that it works with the set up on OSS-Fuzz's side.

sirainen:
I'm thinking that we could make a fuzzer branch here to dovecot github repository, where we could start accepting pull requests. After it's been going on for a while we can then figure out if that works well or if some redesigns are needed. That way we also don't start polluting the master branch's git history too early :) Also we couldn't accept pull requests directly from github to master branch anyway, because the development is done in our internal gitlab that just syncs to github and that would break.

DavidKorczynski:
I like the incremental approach so that sounds good to me.

catenacyber:
> What's the benefit in using clang++ for linking? Oh, or because LIB_FUZZING_ENGINE contains a c++ library?

Exactly, libFuzzer is in C++ cf https://github.com/llvm-mirror/compiler-rt/tree/master/lib/fuzzer

> I pushed to #117 some kludging to make it link fuzz-targets with CXXLINK. Seems to work against my test c++ library but I didn't try against real oss-fuzz.

Cool, with dummy CXX file ;-) I will try it

> I'm thinking that we could make a fuzzer branch here to dovecot github repository

Good idea, but you will need to merge the commits fixing bugs in master in this fuzz branch as well

sirainen:
fuzzer branch now has my pull request merged.

cmouse:
This has now been done.

PR #117, opened 2020-03-23
Fuzzer framework

sirainen:


sirainen:
Accidentally wrote a comment linking to this page to the wrong project, sorry! Deleted already but I guess it'll send email notifications..

src/lib-imap/fuzz-imap-utf7.c
```diff
@@ -0,0 +1,14 @@
+/* Copyright (c) 2020 Dovecot authors, see the included COPYING file */
+
+#include "lib.h"
+#include "fuzzer.h"
+#include "imap-utf7.h"
+
+FUZZ_STR(const char *str)
+{
+	string_t *dest = t_str_new(size);
```
catenacyber: use of `size` feels magic because of the macro

src/lib-test/fuzzer.h
```diff
@@ -0,0 +1,16 @@
...
+#define FUZZER_H
+
+#define FUZZ_DATA(args) \
+	int LLVMFuzzerTestOneInput(args); \
+	int LLVMFuzzerTestOneInput(args) {
+
+#define FUZZ_STR(str_arg) \
+	int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size); \
+	int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) { \
+		str_arg = t_strndup(data, size);
```
catenacyber: Should this not be freed ?

cmouse: t_stuff do  not need to be explicitly freed.

m4/dovecot.m4
```diff
@@ -342,6 +342,19 @@ AC_DEFUN([DC_DOVECOT_HARDENING],[
...
 ])
 
+AC_DEFUN([DC_DOVECOT_FUZZER],[
+        AC_ARG_ENABLE(fuzzer,
+        AS_HELP_STRING([--enable-fuzzer=yes], [Enable clang fuzzer (default: no)]),
+                enable_fuzzer=$enableval,
+                enable_fuzzer=yes)
+	AS_IF([test x$enable_fuzzer = xyes], [
+		CFLAGS="$CFLAGS -fsanitize=fuzzer-no-link"
+		FUZZER_LDFLAGS=-fsanitize=fuzzer
```
catenacyber: This should be made more generic
oss-fuzz uses environment variable `LIB_FUZZING_ENGINE`
On MacOS, this does not work and you have to set `LIB_FUZZING_ENGINE=/path/to/libFuzzer.a`

src/lib-test/fuzzer.h
sirainen: Is LLVMFuzzerTestOneInput running in a loop? I didn't find out how exactly it was called, so I thought maybe it recreates the process between runs. But I guess not?

m4/dovecot.m4
sirainen: It's not very clear to me how these are supposed to work. So clang has libfuzzer and I got the above parameters from it. But then oss-fuzz does something else on top of clang libfuzzer? Or not? Don't I need any of the -fsanitize=fuzzer flags to run with oss-fuzz after all? I don't really understand why clang wants to build sources with this -fsanitize=fuzzer anyway.

src/lib-imap/fuzz-imap-utf7.c
sirainen: Oh, right, I was just copying it from the original test. I didn't intend to have/use hidden variables, although they do exist.. Should rename them at least.

src/lib-test/fuzzer.h
sirainen: Added T_BEGIN .. T_END

m4/dovecot.m4
catenacyber: You can see it this way : 
There are two things :
- fuzz targets : those are the files like fuzz-imap-utf7.c with use of `LLVMFuzzerTestOneInput`
- fuzz drivers : this can be libFuzzer (from clang with `-fsanitize=fuzzer`), but there are others such as AFL... Oss-fuzz uses libFuzzer, AFL and HongFuzz as drivers. That is why they use this generic `LIB_FUZZING_ENGINE`. Basically they are a `main` function calling repeatedly `LLVMFuzzerTestOneInput`...

> Don't I need any of the -fsanitize=fuzzer flags to run with oss-fuzz after all?

These are set by `CFLAGS` in oss-fuzz

src/lib-test/fuzzer.h
catenacyber: > t_stuff do not need to be explicitly freed

Oh... How does it work ?

> Is LLVMFuzzerTestOneInput running in a loop?

Usually yes cf libFuzzer. But you can write a simple driver that runs it only once, or that runs it for all the files in a specific directory...

m4/dovecot.m4
DavidKorczynski: @sirainen  in my PR to OSS-Fuzz where I added the Dovecot project you can see how I only use the CFLAGS and LIB_FUZZING_ENGINE when building the fuzzers, but never actually set variables as OSS-Fuzz does this for you for free https://github.com/google/oss-fuzz/pull/3502/commits/91e5c2eaedf354fb640961dbe96a15286d793b58#diff-422c46e236d34703c859d7d6378624e8

sirainen: OK, so with oss-fuzz you don't need to set any special CFLAGS? I tested with clang that it really does need them or it'll fail with:

ERROR: no interesting inputs were found. Is the code instrumented for coverage? Exiting.

I've now implemented `--enable-fuzzer=clang` for the clang method and `--enable-fuzzer=oss` that only links to LIB_FUZZING_ENGINE. Does this work?

DavidKorczynski: That's correct - the default values that OSS-Fuzz sets the flags to is something like this: `CFLAGS= -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link` 
and `CXXFLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++`

DavidKorczynski: The error you get is because the "fuzzer-no-link" was not used when compiling the object files of the projekt. In general, we should use "-fsanitize=fuzzer-no-link" when compiling and "-fsanitize=fuzzer" when linking

sirainen: Ah, so CFLAGS are set in environment so it overrides the configure-provided CFLAGS. What does LIB_FUZZING_ENGINE expand to then? Does it contain the -fsanitize=fuzzer when linking, or is that added to LDFLAGS environment also, or something?

DavidKorczynski: Yeah, LIB_FUZZING_ENGINE expands to -fsanitize=fuzzer. However, notice this is the default setting and Oss-Fuzz might set the flags a bit different if, for example, we are compiling for a "coverage" build. Then, CFALGS expands to something like `CFLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION  -fprofile-instr-generate -fcoverage-mapping -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument` and `CXXFLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION  -fprofile-instr-generate -fcoverage-mapping -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -stdlib=libc++`

sirainen: Okay, so another attempt: Now there's again only `configure --enable-fuzzer` which sets `CFLAGS=-fsanitize=fuzzer-no-link` but that doesn't matter because oss-fuzz overwrites it anyway. It also uses `-fsanitize=fuzzer` for linking, except if `$LIB_FUZZING_ENGINE` environment is set then that is used for linking instead.

DavidKorczynski: Sounds good to me!

m4/dovecot.m4
```diff
@@ -342,6 +342,26 @@ AC_DEFUN([DC_DOVECOT_HARDENING],[
 	AC_LD_RELRO
 ])
 
+AC_DEFUN([DC_DOVECOT_FUZZER],[
+        AC_ARG_ENABLE(fuzzer,
+        AS_HELP_STRING([--enable-fuzzer=yes], [Enable fuzzer (default: no)]),
```
cmouse: should this be `--with-fuzzer` instead, like clang?

sirainen: I'm not really sure what the main difference is between "enable" and "with". I've been thinking that "with" is used when using some external library for example. This fuzzer can be just about compiler/linker options but then again it could be linking external library via `$LIB_FUZZING_ENGINE`. So from compiler flags point of view it's similar to `--enable-hardening=yes`. But then there's `--with-ioloop` and `--with-notify` which are about choosing which functions to use, even though they're not external libraries.. I wonder if those should have been using "enable" instead.

Anyway, I'm not sure. I could accept either way.

DavidKorczynski: btw. could you provide me with an email where we send the bug reports to? This is needed when merging to OSS-Fuzz (the email will be public in the oss-fuzz Repo)

sirainen: Here: oss-fuzz@open-xchange.com

DavidKorczynski: Perfect, I will fix up the OSS-Fuzz side once this PR is merged.

sirainen: Changed to use `--with-fuzzer` (And hopefully didn't change anything else, because I messed up a few times with git losing changes and had to restore them.)

DavidKorczynski:
I just ran the build within oss-fuzz and can confirm that it works.

m4/dovecot.m4
catenacyber: > Ah, so CFLAGS are set in environment so it overrides the configure-provided CFLAGS.

CFLAGS are indeed set in environment and configure should only _expand_ them (there should be no overriding)

catenacyber:
Seems good to me also

src/lib-imap/Makefile.am
```diff
@@ -96,3 +96,16 @@ check-local:
 	for bin in $(test_programs); do \
 	  if ! $(RUN_TEST) ./$$bin; then exit 1; fi; \
 	done
+
+if USE_FUZZER
+noinst_PROGRAMS += \
+	fuzz-imap-utf7
+
+nodist_EXTRA_fuzz_imap_utf7_SOURCES = force-cxx-linking.cxx
```
cmouse: is this file generated or what? why not put it in EXTRA_SOURCES outside the if block?

src/lib-test/fuzzer.h
```diff
@@ -0,0 +1,16 @@
+#ifndef FUZZER_H
+#define FUZZER_H
+
+#define FUZZ_DATA(args) \
```
cmouse: should these be FUZZ_START_DATA and FUZZ_START_STR to ensure we realize to call FUZZ_END too?

src/lib-test/fuzzer.h
```diff
@@ -0,0 +1,16 @@
...
+	int LLVMFuzzerTestOneInput(args); \
```
cmouse: should this be ((args)) ?

src/lib-test/fuzzer.h
```diff
@@ -0,0 +1,16 @@
...
+	int LLVMFuzzerTestOneInput(args) {
+
+#define FUZZ_STR(str_arg) \
+	int LLVMFuzzerTestOneInput(const uint8_t *_param_data, size_t _param_size); \
+	int LLVMFuzzerTestOneInput(const uint8_t *_param_data, size_t _param_size) { \
+		T_BEGIN { str_arg = t_strndup(_param_data, _param_size);
```
cmouse: should this be ((str_arg)) ?

src/lib-imap/Makefile.am
sirainen: It's a kludgy way to use C++ linking. The file doesn't actually exist. It's recommended in https://www.gnu.org/software/automake/manual/html_node/How-the-Linker-is-Chosen.html

src/lib-test/fuzzer.h
sirainen: No, my idea was that it would be used like `FUZZ_DATA(const uint8_t *data, size_t size)` when binary input is wanted to be tested rather than NUL-terminated string.

src/lib-test/fuzzer.h
sirainen: Again no, that would fail to compile then when using `FUZZ_STR(const char *str)`

src/lib-test/fuzzer.h
sirainen: Yeah, could be clearer. Although for T_BEGIN/END consistency I used FUZZ_BEGIN_*

src/lib-test/fuzzer.h
cmouse: but what's FUZZ_END then?

src/lib-test/fuzzer.h
cmouse: ah, it works like that.

src/lib-imap/Makefile.am
sirainen: Although it would be nicer if this was done across all binaries so it wouldn't have to be repeated for every single fuzzer binary.. Since this is only used for fuzzing, it wouldn't matter if other binaries were also linked with clang++. I'm not sure if there's any way to do that though.

src/lib-imap/fuzz-imap-utf7.c
```diff
@@ -0,0 +1,15 @@
...
+#include "fuzzer.h"
+#include "imap-utf7.h"
+
+FUZZ_BEGIN_STR(const char *str)
+{
+	string_t *dest = t_str_new(32);
+
+	imap_utf8_to_utf7(str, dest);
+	imap_utf7_to_utf8(str, dest);
+}
```
cmouse: just want to point out that these { } are not really needed. feel free to resolve =)

src/lib-imap/Makefile.am
sirainen: I suppose the fuzz*.c sources could be C++ sources directly, but I'm not sure if that causes problems. And could be nice to be able to compile the fuzzers also when C++ wasn't required to be used.

src/lib-imap/fuzz-imap-utf7.c
sirainen: I think they're still useful. Both for humans and also for editors' auto-indenting.

PR #119, opened 2020-03-26
Fuzz target for SMTP server

catenacyber:
How is this @sirainen ?

### DIFF
--- a/src/lib-smtp/fuzz-smtp-server.c
+++ b/src/lib-smtp/fuzz-smtp-server.c
@@ -1,0 +1,51 @@
+/* Copyright (c) 2020 Catena cyber */
+
+#include "lib.h"
+#include "fuzzer.h"
+#include "ioloop.h"
+#include "smtp-server.h"
+
+static void server_connection_destroy(void *context)
+{
+	struct ioloop *ioloop = context;
+	io_loop_stop(ioloop);
+}
+
+static void test_server_continue(void *unused)
+{
+	//instead of simple io_loop_stop so as to free input io
+	io_loop_stop_delayed(current_ioloop);
+}
+
+FUZZ_BEGIN_DATA(const uint8_t *data, size_t size)
+{
+	struct smtp_server_connection *conn;
+	struct smtp_server_settings smtp_server_set = {
+		.max_client_idle_time_msecs = 500,
+		.max_pipelined_commands = 16,
+		.auth_optional = TRUE,
+	};
+	struct smtp_server_callbacks server_callbacks = {
+		.conn_destroy = server_connection_destroy,
+	};
+	struct smtp_server *smtp_server = NULL;
+	struct ioloop *ioloop;
+	int fd;
+	struct timeout *to;
+
+	ioloop = io_loop_create();
+	to = timeout_add_short(0, test_server_continue, NULL);
+	smtp_server = smtp_server_init(&smtp_server_set);
+
+	fd = fuzzer_io_as_fd(data, size);
+	conn = smtp_server_connection_create(smtp_server, fd, fd, NULL, 0,
+					     FALSE, NULL, &server_callbacks, ioloop);
+	smtp_server_connection_start(conn);
+
+	io_loop_run(ioloop);
+
+	smtp_server_deinit(&smtp_server);
+	timeout_remove(&to);
+	io_loop_destroy(&ioloop);
+}
+FUZZ_END
--- a/src/lib-smtp/Makefile.am
+++ b/src/lib-smtp/Makefile.am
@@ -84,7 +84,14 @@ test_programs = \
 
 test_nocheck_programs =
 
-noinst_PROGRAMS = $(test_programs) $(test_nocheck_programs)
+fuzz_programs =
+
+if USE_FUZZER
+fuzz_programs += \
+	fuzz-smtp-server
+endif
+
+noinst_PROGRAMS = $(fuzz_programs) $(test_programs) $(test_nocheck_programs)
 
 EXTRA_DIST = \
 	test-bin/sendmail-exit-1.sh \
@@ -167,6 +174,13 @@ test_smtp_server_errors_SOURCES = test-smtp-server-errors.c
 test_smtp_server_errors_LDFLAGS = -export-dynamic
 test_smtp_server_errors_LDADD = $(test_libs)
 test_smtp_server_errors_DEPENDENCIES = $(test_deps)
+
+nodist_EXTRA_fuzz_smtp_server_SOURCES = force-cxx-linking.cxx
+fuzz_smtp_server_CPPFLAGS = $(FUZZER_CPPFLAGS)
+fuzz_smtp_server_LDFLAGS = $(FUZZER_LDFLAGS)
+fuzz_smtp_server_SOURCES = fuzz-smtp-server.c
+fuzz_smtp_server_LDADD = $(test_libs)
+fuzz_smtp_server_DEPENDENCIES = $(test_deps)
 
 check-local:
 	for bin in $(test_programs); do \


src/lib-smtp/fuzz-smtp-server.c
```diff
@@ -0,0 +1,96 @@
...
+    to = timeout_add_short(0, test_server_continue, NULL);
+    smtp_server = smtp_server_init(&smtp_server_set);
+
+    /* build input as istream out of data and size */
+    sconn = i_new(struct server_connection, 1);
+    pipe(fdi);
+    pipe(fdo);
+    write(fdi[1], data, size);
+    conn = smtp_server_connection_create(smtp_server, fdi[0], fdo[1],
+                                         NULL, 0, FALSE, NULL, &server_callbacks, sconn);
```
sirainen: I wonder if it works to use smtp_server_connection_create_from_streams() instead? That would avoid using the pipe and would be a bit more reliable, especially with large inputs where write to pipe might not write everything.

```
input = i_stream_create_from_data(data, size);
output = o_stream_create_fd(dev_null_fd, IO_BLOCK_SIZE);
conn = smtp_server_connection_create_from_streams(...)
```

src/lib-smtp/fuzz-smtp-server.c
```diff
@@ -0,0 +1,96 @@
...
+    struct timeout *to;
+    struct smtp_server_settings smtp_server_set;
+    static struct smtp_server_callbacks server_callbacks;
+    static struct smtp_server *smtp_server = NULL;
+
+    if (initialized == 0) {
+        lib_init();
+        hostpid_init();
+        initialized = 1;
+    }
```
sirainen: This looks like something that should be put to FUZZ_START_* macros automatically (hostpid_init() is called internally by lib_init() already):

```
diff --git a/src/lib-test/fuzzer.h b/src/lib-test/fuzzer.h
index b0c3d7625a..24ec60998b 100644
--- a/src/lib-test/fuzzer.h
+++ b/src/lib-test/fuzzer.h
@@ -1,13 +1,18 @@
 #ifndef FUZZER_H
 #define FUZZER_H
 
+#define FUZZ_BEGIN_COMMON \
+	if (!lib_is_initialized) lib_init();
+
 #define FUZZ_BEGIN_DATA(args) \
 	int LLVMFuzzerTestOneInput(args); \
-	int LLVMFuzzerTestOneInput(args) {
+	int LLVMFuzzerTestOneInput(args) { \
+		FUZZ_BEGIN_COMMON T_BEGIN {
 
 #define FUZZ_BEGIN_STR(str_arg) \
 	int LLVMFuzzerTestOneInput(const uint8_t *_param_data, size_t _param_size); \
 	int LLVMFuzzerTestOneInput(const uint8_t *_param_data, size_t _param_size) { \
+		FUZZ_BEGIN_COMMON \
 		T_BEGIN { str_arg = t_strndup(_param_data, _param_size);
 
 #define FUZZ_END \
```

src/lib-smtp/fuzz-smtp-server.c
```diff
@@ -0,0 +1,96 @@
...
+    smtp_set->debug = 0;
+}
+
+
+static void test_server_continue(void *dummy) {
+    //instead of simple io_loop_stop so as to free input io
+    io_loop_stop_delayed(ioloop_fuzz);
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
```
sirainen: this should be:

```
FUZZ_START_DATA(const uint8_t *data, size_t size)
{
```

(I just noticed this is buggy with the current committed version, but fixed by the patch in the other commit)

src/lib-smtp/fuzz-smtp-server.c
```diff
@@ -0,0 +1,96 @@
...
+
+static void
+test_server_defaults(struct smtp_server_settings *smtp_set)
+{
+    /* server settings */
+    i_zero(smtp_set);
+    smtp_set->max_client_idle_time_msecs = 500;
+    smtp_set->max_pipelined_commands = 16;
+    smtp_set->auth_optional = TRUE;
+    smtp_set->debug = 0;
```
sirainen: debug = FALSE .. or just don't set it, since it's the default anyway

src/lib-smtp/fuzz-smtp-server.c
sirainen: I added this to https://github.com/dovecot/core/pull/121

src/lib-smtp/fuzz-smtp-server.c
sirainen: Oh well, looks like it doesn't work. It really wants fds to exist.

src/lib-smtp/fuzz-smtp-server.c
```diff
@@ -0,0 +1,96 @@
...
+    server_callbacks.conn_destroy = server_connection_destroy;
+    ioloop_fuzz = io_loop_create();
+    to = timeout_add_short(0, test_server_continue, NULL);
+    smtp_server = smtp_server_init(&smtp_server_set);
+
+    /* build input as istream out of data and size */
+    sconn = i_new(struct server_connection, 1);
+    pipe(fdi);
+    pipe(fdo);
+    write(fdi[1], data, size);
```
sirainen: These are giving compiler warnings. Should check the return value, e.g.:

```
if (pipe(fdi) < 0) i_fatal("pipe() failed: %m");
if (pipe(fdo) < 0) i_fatal("pipe() failed: %m");
if (write(fdi[1], data, size) < 0) i_fatal("write() failed: %m");
```

src/lib-smtp/fuzz-smtp-server.c
```diff
@@ -0,0 +1,96 @@
...
+    smtp_server_connection_start(conn);
+
+    io_loop_run(ioloop_fuzz);
+
+    smtp_server_deinit(&smtp_server);
+    timeout_remove(&to);
+    io_loop_destroy(&ioloop_fuzz);
+
+    close(fdi[1]);
+    close(fdo[0]);
```
sirainen: And these can be i_close_fd(&fdi[1]); which checks the return values internally

sirainen:
I wanted to make this a bit easier and more reliable since other fuzzing targets will likely have similar code. So based on https://github.com/dovecot/core/pull/121 here's a simplified version of this test:

```
/* Copyright (c) 2020 Catena cyber */

#include "lib.h"
#include "fuzzer.h"
#include "ioloop.h"
#include "smtp-server.h"

static void server_connection_destroy(void *context)
{
	struct ioloop *ioloop = context;
	io_loop_stop(ioloop);
}

FUZZ_BEGIN_DATA(const uint8_t *data, size_t size)
{
	struct smtp_server_connection *conn;
	struct smtp_server_settings smtp_server_set = {
		.max_client_idle_time_msecs = 500,
		.max_pipelined_commands = 16,
		.auth_optional = TRUE,
	};
	struct smtp_server_callbacks server_callbacks = {
		.conn_destroy = server_connection_destroy,
	};
	struct smtp_server *smtp_server = NULL;
	struct ioloop *ioloop;
	int fd;

	ioloop = io_loop_create();
	smtp_server = smtp_server_init(&smtp_server_set);

	fd = fuzzer_io_as_fd(data, size);
	conn = smtp_server_connection_create(smtp_server, fd, fd, NULL, 0,
					     FALSE, NULL, &server_callbacks, ioloop);
	smtp_server_connection_start(conn);

	io_loop_run(ioloop);

	smtp_server_deinit(&smtp_server);
	io_loop_destroy(&ioloop);
}
FUZZ_END
```

src/lib-smtp/fuzz-smtp-server.c
catenacyber: Indeed, this is a leftover

src/lib-smtp/fuzz-smtp-server.c
catenacyber: Indeed, I has tried with `smtp_server_connection_create_from_streams` without success as fds were required.

I found that `pipe` are the most socket-like object without needing another process, but you are welcome if you find a better way to do it :-)

catenacyber:
Thanks for the review.

So, I should rebase this when #121 gets merged and use the modified code as you proposed, right ?

sirainen:
Yes, please rebase. #121 is merged now. Although I'm not 100% sure if my version of the code sees all the input correctly (especially not truncated). I think it does, but I didn't verify.

catenacyber:
I have just rebased, used and modified your code and force-pushed

After testing, it seems to me that we need the call to `timeout_add_short`

catenacyber:
I saw that some fixes were merged.
So, can I friendly ping on this ?

Otherwise, are there specific parts of the code you would like to be fuzzed ?
cc @cmouse

catenacyber:
I just rebased on top of master, and fixed one simple conflict

sirainen:
It's still saying "This branch cannot be rebased due to conflicts" - could you rebase it against the "fuzzer" branch instead of master? That should fix it.

catenacyber:
Thanks @sirainen rebase on the right `fuzzer` branch

PR #120, opened 2020-03-26
Fix bugs in smtp server

catenacyber:
`smtp_server_cmd_bdat` calls `smtp_command_parse_data_with_size` with first parameter being `conn->smtp_parser` which may be assigned to NULL by the previous call to `smtp_server_reply` ending up calling `smtp_server_connection_terminate` with the condition `if conn->bad_counter > conn->set.max_bad_commands`

cf https://hackerone.com/bugs?report_id=831290

The assert needs to be extended because `smtp_server_cmd_helo_run` can call `smtp_server_cmd_ehlo_reply_default`

### DIFF
--- a/src/lib-smtp/smtp-server-cmd-data.c
+++ b/src/lib-smtp/smtp-server-cmd-data.c
@@ -654,7 +654,7 @@ void smtp_server_cmd_bdat(struct smtp_server_cmd_ctx *cmd,
 		}
 	}
 
-	if (ret > 0 || size > 0) {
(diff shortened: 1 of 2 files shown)


catenacyber:
This bug is still present and got found by oss-fuzz https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23508

catenacyber:
And https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23510 for the assert

cmouse:
Fixed via 896f120709666e37955510226475c894b5b95b99

PR #121, opened 2020-03-26
FUZZ_START_* fixes

sirainen:


sirainen:
I wonder if FUZZ_BEGIN_FD() macro would also be useful to avoid having to call fuzzer_io_as_fd()..

PR #122, opened 2020-03-31
lib-compression: fix length checks for zlib header

catenacyber:
Buffer over-read with input such as `\x1f\x8b\x01Dovec/\x01\xbf`


cmouse:
Ticket ref DOP-1763

cmouse:
merged externally.

cmouse:
commit 959e958c

PR #123, opened 2020-04-02
fix one byte overread in parse_angle_addr

catenacyber:
Cf https://hackerone.com/bugs?report_id=836045

Found by fuzzing `message_address_parse_path`

For now, binaries are safe as 
> All of the callers to message_address_parse_path() pass it a pointer to a NUL-terminated string.


cmouse:
we merged our own version. thank you for the effort!

PR #124, opened 2020-04-22
Switch default minimum supported TLS version to v1.2

mrueg:
This PR switches the default minimum to TLSv1.2 as recommended in https://blog.pcisecuritystandards.org/migrating-from-ssl-and-early-tls


dciancu:
+1 for adding TLSv1.3 to supported protocols in Dovecot code, although openssl supports it, we cannot use it due to the check in Dovecot code.

src/config/old-set-parser.c
```diff
@@ -171,7 +171,7 @@ static int ssl_protocols_to_min_protocol(const char *ssl_protocols,
 					 const char **error_r)
 {
 	static const char *protocol_versions[] = {
-		"SSLv3", "TLSv1", "TLSv1.1", "TLSv1.2",
+		"SSLv3", "TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3",
```
darix, line 174: This change alone is not enough to allow "TLSv1.3" only servers.

mrueg: Yes it doesn't add TLSv1.3 per se. I looked through the code, config and documentation for occurrences that did not include it and added it there.

darix: https://github.com/dovecot/core/pull/126

cmouse:
Fixed via 096b7122b9068ed34224a52c067437107f6f72ab

PR #125, opened 2020-04-23
Fix haproxy PROXY protocol LOCAL type connection behavior when header includes address information or TLVs

alexat:
PROXY protocol LOCAL type headers can actually include supplementary information (addresses, TLVs) which need to be skipped to correctly handle the header. Fixes Invalid TLV bug with haproxy 2.0.14 service checks.

More info here: https://github.com/haproxy/haproxy/issues/511
And here: https://www.mail-archive.com/haproxy@formilux.org/msg36890.html


cmouse:
Hi! Thank you for the pull request, we'll look into it.

cmouse:
Tracking as DOV-4319

cmouse:
Merged as f753ea9b31ae098a9435a67de69cbb05e78bbb84

PR #127, opened 2020-05-24
Document command/args destination format to sync/backup.

FGasper:
Per a recent mailing list discussion thread, I noted that the doveadm sync/backup format where _destination_ is a command and arguments is not mentioned in the relevant man page.

This intends to correct for that. I wanted to include Sami’s sample command:
```
doveadm backup -u john -R ssh sshuser@remote "sudo /usr/bin/doveadm dsync-server -u john"
```
… as an example, but then I realized that `dsync-server` doesn’t appear to be documented, either. :(


sirainen:
merged

PR #128, opened 2020-05-24
systemd integration: notify service manager when ready

nmartensen:
With Type=simple or Type=forking, systemd does not really know when the
service is ready to accept connections and might start depending
services too early. Use Type=notify to explicitly tell the service
manager when the service is ready.

For a real problem caused by assuming readiness too early, please see
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=951722

For the meaning of the service type and details of the readiness
protocol, see also the following links:
https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=
https://www.freedesktop.org/software/systemd/man/sd_notify.html

As discussed in the last link, more elaborate state notifications are
possible. This patch only implements the most basic part.

Original patch prepared by Michael Biebl, with slight modification.

A second commit fixes an obvious typo in Makefile.am. I believe
the brackets are needed as well?

### DIFF
--- a/src/master/main.c
+++ b/src/master/main.c
@@ -26,6 +26,9 @@
 #include "service-process.h"
 #include "service-log.h"
 #include "dovecot-version.h"
+#ifdef HAVE_SYSTEMD
+#  include "sd-daemon.h"
+#endif
 
 #include <stdio.h>
 #include <unistd.h>
@@ -544,6 +547,9 @@ static void main_init(const struct master_settings *set)
 	master_clients_init();
 
 	services_monitor_start(services);
+#ifdef HAVE_SYSTEMD
+	sd_notify(0, "READY=1");
+#endif
 	startup_finished = TRUE;
 }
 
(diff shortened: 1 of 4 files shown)


src/master/main.c
```diff
@@ -26,6 +26,9 @@
 #include "service-process.h"
 #include "service-log.h"
 #include "dovecot-version.h"
+#ifdef HAVE_SYSTEMD
+#include "sd-daemon.h"
```
sirainen, line 30: For coding style consistency would be nice if this has two space indentation: `#  include`

sirainen:
Tracking internally in DOP-1904

src/master/main.c
sirainen: Of course github split the line exactly at the wrong place, at least in my browser.. To clarify:

```
#  include "sd-daemon.h"
```

nmartensen: done

cmouse:
Merged externally.

PR #129, opened 2020-05-27
Prevent doveadm from sending extra \n-\n on failure.

FGasper:
This fixes an issue I’ve reported on the mailing list; however, I’m not sure if fixing this would break existing clients.

It may be better just to update the documentation, but I thought I’d send this PR in regardless.

### DIFF
--- a/src/doveadm/client-connection-tcp.c
+++ b/src/doveadm/client-connection-tcp.c
@@ -55,6 +55,8 @@ static failure_callback_t *orig_info_callback, *orig_debug_callback = NULL;
 
 static bool log_recursing = FALSE;
 
+static bool printed_response = FALSE;
+
 static void ATTR_FORMAT(2, 0)
 doveadm_server_log_handler(const struct failure_context *ctx,
 			   const char *format, va_list args)
@@ -169,6 +171,8 @@ doveadm_cmd_server_post(struct client_connection_tcp *conn, const char *cmd_name
 		i_error("BUG: Command '%s' returned unknown error code %d",
 			cmd_name, doveadm_exit_code);
 	}
+
+	printed_response = TRUE;
 }
 
 static void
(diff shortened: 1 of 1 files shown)


FGasper:
Mailing list thread:
https://dovecot.org/pipermail/dovecot/2020-May/119066.html

sirainen:
Hmm. This has been weirdly long in here without any responses. I don't think this affects the normal doveadm client usage, but it does break when trying to use doveadm with wildcard users because the replies can't be matched. Tracking in DOP-2453.

Also here's a simpler fix:

```
diff --git a/src/doveadm/client-connection-tcp.c b/src/doveadm/client-connection-tcp.c
index 4ec3dc7a3c..55c5a28130 100644
--- a/src/doveadm/client-connection-tcp.c
+++ b/src/doveadm/client-connection-tcp.c
@@ -360,7 +360,7 @@ static int doveadm_cmd_handle(struct client_connection_tcp *conn,
        /* clear all headers */
        doveadm_print_deinit();
        doveadm_print_init(DOVEADM_PRINT_TYPE_SERVER);
-       return doveadm_exit_code == 0 ? 0 : -1;
+       return 0;
 }

 static bool client_handle_command(struct client_connection_tcp *conn,
```

FGasper:
@sirainen Thank you for the update. In the meantime we’ve accommodated the quirk in our logic, so we’ll keep an eye on DOP-2453.

FGasper:
Actually, do you have a reference to that issue? Or is that an internal reference only?

sirainen:
Sorry, it's internal only. Anyway, the fix is likely going to v2.3.17 earliest.

cmouse:
merged as 9ea5d36197d9193a7a9d49c99cbb12148e56e404

PR #131, opened 2020-07-27
OBJECTID IMAP extension

filiphanes:
https://tools.ietf.org/html/rfc8474
THREADID == EMAILID for now, as it is allowed by RFC


cmouse:
Hi! Thank you for your merge request, we'll look into this.

src/imap/imap-fetch.c
```diff
@@ -916,6 +916,66 @@ static bool imap_fetch_savedate_init(struct imap_fetch_init_context *ctx)
...
+
+static int fetch_threadid(struct imap_fetch_context *ctx, struct mail *mail,
+		      void *context ATTR_UNUSED)
+{
+	const char *value;
+
+	if (mail_get_special(mail, MAIL_FETCH_GUID, &value) < 0)
+		return -1;
+
+	str_printfa(ctx->state.cur_str, "THREADID (%s) ", value);
```
sirainen, line 947: I think this should be NIL until we actually support it.

src/imap/imap-fetch.c
```diff
@@ -916,6 +916,66 @@ static bool imap_fetch_savedate_init(struct imap_fetch_init_context *ctx)
...
+static int fetch_mailboxid(struct imap_fetch_context *ctx, struct mail *mail,
+			   void *context ATTR_UNUSED)
+{
+	struct mailbox_metadata metadata;
+
+	if (mailbox_get_metadata(mail->box, MAILBOX_METADATA_GUID, &metadata) < 0)
+		return -1;
+
+	str_printfa(ctx->state.cur_str, "MAILBOXID (%s) ",
+			guid_128_to_string(metadata.guid));
```
sirainen, line 968: There's no MAILBOXID for FETCH, only STATUS.

src/lib-storage/mail-search-register-human.c
```diff
@@ -172,6 +172,11 @@ static const struct mail_search_register_arg human_register_args[] = {
 	{ "LARGER", human_search_larger },
 	{ "SMALLER", human_search_smaller },
 
+	/* OBJECTID extension: */
+	{ "EMAILID", human_search_guid },
+	{ "THREADID", human_search_guid },
+	{ "MAILBOXID", human_search_mailbox_guid },
```
sirainen, line 178: Also no MAILBOXID for SEARCH.

src/imap/imap-fetch.c
filiphanes: ok, any ideas how could by THREADID implemented?

src/imap/imap-fetch.c
filiphanes: yes, but there is already X-MAILBOX, this could be more standardized name and usable for clients when using virtual \All mailbox

src/lib-storage/mail-search-register-human.c
filiphanes: same as MAILBOXID for FETCH, I am planning to use it in our webmail client

cmouse:
Please reopen against main.

src/imap/imap-fetch.c
sirainen: And as for answer to this: With lots of pain and effort. My thinking has been to create one or two new index files where you can get msgid <-> threadid mappings. I've a more detailed (but not finished) plan written out (DOP-768 in our internal Jira).

sirainen:
DOP-639 - Looks like we also have a similar conflicting patch by Stephan.

PR #134, opened 2020-08-17
Fix APOP test failure on 32-bit systems

pghmcfc:
The challenge timestamp is read as a hex number in `mech_apop_auth_initial()` so it should be written as hex, not decimal.

Also fix compiler warnings for 32-bit architectures.


cmouse:
Hi! Thank you for your pull request, we already have this in our own review queue though internally.

cmouse:
Fixed with a744f282947907c794bd23838775b9124c90b11d

PR #138, opened 2020-09-28
fix Buffer Over-read

ihsinme:
in the source file "mail-html2text.c" in the function "parse_tag_name" there is an incorrect check if (size> = 10 && i_memcasecmp (data, "blockquote", 10) == 0 && (HTML_WHITESPACE (data [10]) || data [10] == '>')).

the specified check results in access outside the allocated memory.

### DIFF
--- a/src/lib-mail/mail-html2text.c
+++ b/src/lib-mail/mail-html2text.c
@@ -87,7 +87,7 @@ parse_tag_name(struct mail_html2text *ht,
 		return 8 + 1;
 	}
 
-	if (size >= 10 && i_memcasecmp(data, "blockquote", 10) == 0 &&
+	if (size >= 11 && i_memcasecmp(data, "blockquote", 10) == 0 &&
 	    (HTML_WHITESPACE(data[10]) || data[10] == '>')) {
 		ht->quote_level++;
 		ht->state = HTML_STATE_TAG;


src/lib-mail/mail-html2text.c
```diff
@@ -87,7 +87,7 @@ parse_tag_name(struct mail_html2text *ht,
 		return 8 + 1;
 	}
 
-	if (size >= 10 && i_memcasecmp(data, "blockquote", 10) == 0 &&
+	if (size >= 11 && i_memcasecmp(data, "blockquote", 10) == 0 &&
```
jeffpc, line 90: I wonder if this code is making an assumption about nul-termination.

ihsinme: if you look at similar checks above, it will become clear that this is a simple slip of the tongue.

jeffpc: The only caller of this function is parse_data(), which passes in the size `size - i - 1`.  Since i >= 0, the size passed to parse_tag_name() will always be < the actual allocated buffer size.

With that said, keeping that sort of assumption would be a bad idea. Based on the code higher up for script and style, I think your patch is good. I'll post it for internal review & CI.

Thanks.

ihsinme: Good afternoon.
I hope you are doing well.
How are you doing on this PL?

The project has no approved PR for a long time.
Perhaps I missed some sad news, or this is the normal state of affairs for the project.

cmouse: Sorry these take very long for us. I'll see what we can do.

sirainen: Also, we merge everything in our internal gitlab and close the pull requests. So there are no merged pull requests in github.

ihsinme: Thank you for your prompt response.

PR #140, opened 2020-10-08
man page doveadm-search-query

raphael-walther:
According to RFC3501 and a test on dovecot 2.3.11.3 (502c39a), doveadm search with search key TEXT search in the header and body of the message. This is not clear in the man page.

Command:
doveadm search -u user@example.org INBOX TEXT test

but man page doveadm-search-query said:
TEXT string
Matches messages, which contain string in the message body.

https://tools.ietf.org/html/rfc3501#section-6.4.4
TEXT
Messages that contain the specified string in the header or
body of the message.


sirainen:
Thanks! Tracking internally in Jira DOV-4145.

sirainen:
Thanks, merged.

PR #142, opened 2020-10-12
Message-ID fails DKIM, Message-Id is better

OzBob:
https://tools.ietf.org/html/rfc5321 mentions addition of a message-id.
Adding a Message-ID if missing has a side effect. Processing the message adds the missing Message-ID, but SMTP servers then add an incorrect DKIM signature, causing outlook, hotmail and yahoo (tested) to add DKIM:fail header, putting the message into Junk or Spam folder.
Adding a Message-Id rather than a Message-ID by the client or SMTP server will work around this feature.

### DIFF
--- a/src/doveadm/doveadm-mail-deduplicate.c
+++ b/src/doveadm/doveadm-mail-deduplicate.c
@@ -90,7 +90,7 @@ cmd_deduplicate_box(struct doveadm_mail_cmd_context *_ctx,
 	hash_table_create(&hash, pool, 0, str_hash, strcmp);
 	while (doveadm_mail_iter_next(iter, &mail)) {
 		if (ctx->by_msgid) {
-			if (mail_get_first_header(mail, "Message-ID", &key) < 0) {
+			if (mail_get_first_header(mail, "Message-Id", &key) < 0) {
 				errstr = mailbox_get_last_internal_error(mail->box, &error);
 				if (error == MAIL_ERROR_NOTFOUND)
 					continue;
--- a/src/lda/main.c
+++ b/src/lda/main.c
@@ -45,7 +45,7 @@ struct event_category event_category_lda = {
 };
 
 static const char *wanted_headers[] = {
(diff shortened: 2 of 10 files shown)


sirainen:
Are you sure this isn't a security hole in Outlook/Hotmail/Yahoo DKIM validation? Since message headers are supposed to be case-insensitive..

sirainen:
Also, I'm not quite sure what exactly is the problem being fixed here? What part in Dovecot causes addition of Message-ID header that ends up in Hotmail/etc and gets invalid DKIM signature? In the patch I can only see mail rejection headers being modified, the others are for read-only access. Shouldn't the rejection mail have its own whole new DKIM signature?

I could understand the issue happening for forwarded emails, but I don't think the patch changes that. Also I'm not sure if Dovecot adds missing Message-ID headers to forwarded emails anyway?

sirainen:
Or is it your local DKIM module that signs outgoing messages wrong when it sees Message-ID rather than Message-Id?

OzBob:
The Message-ID header is an definite issue, but I am searching where to put the fix. 

Is dovecot case insensitive and works for Message-ID and message-id? 
Dovecot does indeed deal with Message-ID here:
https://github.com/dovecot/core/blob/master/src/lib-lda/mail-deliver.c#L121

RFC mentions message-id


> Are you sure this isn't a security hole in Outlook/Hotmail/Yahoo DKIM validation? Since message headers are supposed to be case-insensitive..

I have attempted to seek information from Outlook and Yahoo and Google support but their Operations team have said their DKIM validation is OK.

My ISP says it's an MTA's responsibility to know which headers to add.  I have asked for URL's to support documentation or RFCs or anything that indicate that MTA's need to add a Message-Id.

There may indeed be an issue with the CPANEL/WHM DKIM module.  Can you find the source code on that?  I am still searching.

OzBob:
Is dovecot case insensitive and correctly copies Message-ID and message-id headers?
https://github.com/dovecot/core/blob/master/src/lib-lda/mail-deliver.c#L121

PR #143, opened 2020-10-22
fix incorrect type conversion

ihsinme:
the used transformation is incorrect. first, the sum will be calculated in a basic form, and only then there will be a senseless conversion of the result to an increasing type.

to fix this, I suggest to just move one parenthesis.


sirainen:
The point of the cast is to prevent warnings with 32bit time_t: warning: comparison of integer expressions of different signedness

Your change brings back this warning.

sirainen:
The calculation with unsigned int works until around year 2106. There's a bunch of timestamp code in Dovecot that will break around that time, but I'm expecting something will have replaced Dovecot by then.

PR #144, opened 2020-11-10
maildirlock: Fix and add argument release-timeout

gurnec:
This is one way to fix the currently-broken maildirlock utility as discussed [here](http://dovecot.2317879.n4.nabble.com/maildirlock-time-unit-td65122.html) and [here](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=932253). To summarize its current state:

```sh
$ /usr/lib/dovecot/maildirlock
Usage: maildirlock <path> <timeout>
 - SIGTERM will release the lock.
$ pid=$(/usr/lib/dovecot/maildirlock /tmp 10)
$ Panic: BUG: No IOs or timeouts set. Not waiting for infinity.
Error: Raw backtrace: ...
```

This introduces a timeout on the maximum time a maildir lock can be held by the utility:

```sh
$ /usr/local/lib/dovecot/maildirlock
Usage: maildirlock <path> <acquire-timeout> [<release-timeout>]
 - if unspecified, release-timeout will be the same as acquire-timeout;
   SIGTERM will release the lock.
$ pid=$(/usr/local/lib/dovecot/maildirlock /tmp 10 10); echo $pid; sleep 6; ps -p $pid; sleep 6; ps -p $pid
1515383
    PID TTY          TIME CMD
1515383 pts/1    00:00:00 maildirlock
    PID TTY          TIME CMD
```

`<release-timeout>` is optional for backwards compatibility. When not specified, it's set to the same value as `<acquire-timeout>`—I'm not sure I like this, but it was the best I could come up with for a default.

I'm not claiming this is the best way to fix this, or even a good way, only that it works-for-me<sup>tm</sup> and may for others too. FYI I use maildirlock for backup purposes.

### DIFF
--- a/src/util/maildirlock.c
+++ b/src/util/maildirlock.c
@@ -23,6 +23,11 @@ static void sig_die(const siginfo_t *si ATTR_UNUSED, void *context ATTR_UNUSED)
 	io_loop_stop(ioloop);
 }
 
+static void release_timeout_cb(void *context ATTR_UNUSED)
+{
+	io_loop_stop(ioloop);
+}
+
 static int maildir_lock(const char *path, unsigned int timeout,
 			struct dotlock **dotlock_r)
 {
@@ -37,14 +42,16 @@ static int maildir_lock(const char *path, unsigned int timeout,
 int main(int argc, const char *argv[])
 {
 	struct dotlock *dotlock;
-	unsigned int timeout;
+	struct timeout *timeout;
+	unsigned acquire_timeout_s, release_timeout_s, release_timeout_ms;
 	pid_t pid;
 	int fd[2], ret;
 	char c;
 
-	if (argc != 3) {
-		fprintf(stderr, "Usage: maildirlock <path> <timeout>\n"
-			" - SIGTERM will release the lock.\n");
+	if (argc < 3 || argc > 4) {
+		fprintf(stderr, "Usage: maildirlock <path> <acquire-timeout> [<release-timeout>]\n"
+			" - if unspecified, release-timeout will be the same as acquire-timeout;\n"
+			"   SIGTERM will release the lock.\n");
 		return 1;
 	}
 	if (pipe(fd) != 0) {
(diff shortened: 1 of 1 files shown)


sirainen:
There's also a simpler patch possible nowadays:

```
diff --git a/src/util/maildirlock.c b/src/util/maildirlock.c
index 89fbacc3d5..9fa880af14 100644
--- a/src/util/maildirlock.c
+++ b/src/util/maildirlock.c
@@ -63,7 +63,7 @@ int main(int argc, const char *argv[])
        lib_signals_init();
        ioloop = io_loop_create();
        lib_signals_set_handler(SIGINT, LIBSIG_FLAG_DELAYED, sig_die, NULL);
-       lib_signals_set_handler(SIGTERM, LIBSIG_FLAG_DELAYED, sig_die, NULL);
+       lib_signals_set_handler(SIGTERM, LIBSIG_FLAG_DELAYED | LIBSIG_FLAG_EXPECTED, sig_die, NULL);

        if (pid != 0) {
                i_close_fd(&fd[1]);
```

But we're likely removing maildirlock entirely. Although that has gotten stuck as well.

gurnec:
I couldn't find `LIBSIG_FLAG_EXPECTED`, but you did lead me to this, thank you:

```c
/* Indicate whether signals are expected for the indicated delayed handler. When
   signals are expected, the io for delayed handlers will be allowed to wait
   alone on the ioloop.  */
void lib_signals_set_expected(int signo, bool expected,
			      signal_handler_t *handler, void *context);
```

> But we're likely removing maildirlock entirely.

I think that means that there will no longer be a way to perform certain types of backups which are all three of supported, safe, and efficient, correct? `backup` and `sync` are supported and safe, but require additional copying/space which `maildirlock` does not.

I don't mean to sound like I'm complaining, just trying to understand what's down the road.

Tofandel:
Any activity there? Why would maildirlock be removed? Trying to understand the logic there as it's quite useful (well when it wasn't broken) to perform mail operations like compression. So what's the alternative?

cmouse:
This tool has now been migrated into https://github.com/dovecot/tools, see https://github.com/dovecot/tools/README.maildirlock

PR #145, opened 2020-11-11
doveadm-mail-server: configure TLS on non proxied request

Feandil:
When remotely calling doveadm through the `socket-path` argument (for example `doveadm mailbox list -u vtestuser -S ${host}:{$port}`, without doveadm_port configured (not on a proxy host), we still need to configure TLS in case the remote endpoint uses TLS.

### DIFF
--- a/src/doveadm/doveadm-mail-server.c
+++ b/src/doveadm/doveadm-mail-server.c
@@ -196,13 +196,13 @@ doveadm_mail_server_user_get_host(struct doveadm_mail_cmd_context *ctx,
 	*host_r = ctx->set->doveadm_socket_path;
 	*port_r = ctx->set->doveadm_port;
 
-	if (ctx->set->doveadm_port == 0)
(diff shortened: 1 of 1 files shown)


cmouse:
Fixed with 6d803d99827acc5def479ce93563df8bc3419c33

Feandil:
Thanks @cmouse for the update. A bit sad that no attribution was kept (the content of the fix is different since the code changed in the mean time, but the logic is the same, moving the ssl settings up or the return 0 down), but well...

sirainen:
Sorry, it got fixed as a part of massive changes to the proxying code I was doing. I hadn't even realized this pull request existed at that time. Unfortunately these pull requests are too often left to rot for a long time. :(

Feandil:
No worries. Thanks for the fix in any case :)
A bit sad that it had been left to rot indeed, I guess I should have insisted more ;)

PR #146, opened 2020-11-16
db-lua: Fix invalid stack empty check

mar-v-in:
3e11b97 introduced assertions to ensure the lua stack is empty after execution.
In `auth_lua_call_password_verify` this check was inserted *before* popping the error message from the stack, ultimately causing the assertion to fail every time the lua script has a runtime error, leading to process crashing.


cmouse:
Thank you for your merge request, we'll look into this.

cmouse:
Merged as 2c57a92661b004b8e375f95c1bda0bbddac4395f

PR #149, opened 2021-01-04
Fix timeval_cmp_margin for 32-bit systems

pghmcfc:
The test suite compares times with seconds values of -INT_MAX and INT_MAX. The result of this comparison does not fit in a value of type int and so the test suite fails on 32-bit systems where time_t is an int. To fix this, calculations on seconds values are done using long long integers.


sirainen:
Thanks! Looks fine to me. Tracking internally in DOP-2213.

cmouse:
Merged as 2cc1feca9087651902a5ea3cda021c8a0b3217ce

PR #150, opened 2021-01-05
Add ability to set "default" SASL EXTERNAL username

BryanJacobs:
Please excuse my thorough explanation below; I want to be sure this issue is well-understood so the right decision can be made about where to go.

Today, Dovecot supports a variety of SASL mechanisms. Among them are a few oddball special auth methods. SASL EXTERNAL is one of those. SASL EXTERNAL is designed for situations where the authentication is, well, external to the application. In Dovecot's case, this usually denotes that a user has authenticated using an X.509 certificate (also called "mTLS", "TLS mutual authentication", or "SSL certificate auth").

However, Dovecot is not *just* a POP/IMAP server. It's also a SASL authenticator available to other systems.

Let's say you have a goal of getting rid of password-based auth methods, and you don't want to run or rent a Kerberos KDC. These two constraints together eliminate all IMAP authentication methods save one: SASL EXTERNAL. So you choose SASL EXTERNAL with TLS certificates.

Dovecot handles these beautifully. A client connects, presents the certificate, you set Dovecot's auth_ssl_require_client_cert and auth_ssl_username_from_cert flags, and you're good to go.

Now you want to add SMTP to the equation. You install Postfix, the world's most common standalone SMTP MTA. You configure Postfix to also accept TLS certificates instead of usernames and passwords. This works fine: Postfix doesn't do any SASL auth because it's accepting users as authenticated based on the presented certificates' validity.

Now let's say you have an Android device. You install K-9 mail, a very popular email application. K-9 mail supports certificate auth, yay! So you connect K-9 mail to Dovecot for IMAP mail access, and you try to connect K-9 mail to Postfix for SMTP mail delivery, and... it doesn't work.

K-9 mail, after it performs a certificate-based exchange with an SMTP server, sends an `AUTH EXTERNAL` SMTP stanza to the server. When Postfix receives such a stanza, it calls its configured SASL library. You have two choices of SASL library to use: Dovecot and Cyrus-SASL. You use Dovecot.

Postfix passes the user's authentication credentials to Dovecot. Since the user didn't actually log in, these are a username of NULL and a password of NULL. The certificate the user presented to Postfix isn't forwarded to Dovecot, of course.

Dovecot rejects the authentication with a weird error: "username not known". This is because the SASL external mechanism in Dovecot has an explicit check that disallows NULL usernames.

Let's review the problem:
- K-9 mail on Android always issues an AUTH EXTERNAL command after connecting with a TLS cert, and cannot be configured to do otherwise
- Postfix always calls its SASL library when receiving an AUTH command, and cannot be configured to do otherwise, and cannot be configured to proceed if the SASL library rejects the auth attempt
- Postfix never passes a username to its SASL library when so called by K-9 mail
- Dovecot always throws an exception when it receives a null username, even where that username has no relevance to the authn/authz process

Now, the blame could be placed on any of the parts in this system. "We should fix K-9 mail". "We should fix Postfix". "We should fix Dovecot". None of those folks seems to be moving particularly quickly to fix their parts of the world.

So I'm here to fix Dovecot. In this pull request, I've done all the work of making Dovecot accept an additional configuration parameter, called "auth_external_default_username". When this is set, AND Dovecot receives a NULL username via SASL EXTERNAL, Dovecot pretends it actually received the configured username instead. This allows auth_ssl_username_from_cert to work as intended, and also lets any user-given external authn/authz systems detect that a null username was furnished (by configuring a not-normally-valid name as the default and then special-casing that name in their system).

I've tested that this change unbreaks the above scenario. I've written unit tests. I've based the change on the latest Dovecot master. I've tried to be as minimally invasive as possible. Example of a working config with this change:

```
ssl_verify_client_cert = yes
auth_ssl_username_from_cert = yes
ssl_cert_username_field = emailAddress
auth_external_default_username = saslexternalfakeuser

auth_mechanisms = external

protocol smtp {
        auth_ssl_require_client_cert = no
        auth_ssl_username_from_cert = no
        auth_mechanisms = external
        userdb {
          driver = static
          args = uid=9000 gid=9000 home=/dev/null user=saslexternalfakeuser password=""
          result_failure = return-ok
          result_internalfail = return-ok
        }
        passdb {
          driver = static
          result_failure = return-ok
          result_internalfail = return-ok
          default_fields = user=saslexternalfakeuser
        }
}
```

This will, for SMTP only (ie Postfix), ignore basically everything and return "auth accepted" on any use of the EXTERNAL method. IMAP can be configured completely separately and will do real validation of the user (but probably not the password...).

There is a related (but different) pull request, https://github.com/dovecot/core/pull/64 , which has sat for more than two years since its last activity. That pull request would complement this one, I think. I don't know why it's been there for two years. It looks fine to me. It doesn't solve this problem because it leaves Dovecot failing on a null username.

If you need me to do something more here, all you need to do is let me know.

Thanks for building such a nice IMAP server/POP server/SASL library!


cmouse:
Or you could try dovecot submission server to deal with your users sending mail, which supports SASL EXTERNAL better.

BryanJacobs:
Thunderbird (not mentioned in my explanation above) will not send an AUTH command when using EXTERNAL.

Dovecot's submission server responds with "5.7.0 Authentication required" if you deliver a "MAIL FROM" command over a TLS-secured connection without an explicit SMTP AUTH command. It does this without even trying to contact the MTA behind it. So switching to using Dovecot as an MTA intermediary would fix K-9 mail and break Thunderbird.

The whole thing is a giant mess. Please help out by making Dovecot more flexible, so it can accommodate the variety of behaviours in the wild.

If you have a different preferred set of changes to make this software work together, please give me that constructive feedback so I can make it happen. This pull request was the smallest change that made everything fit.

cmouse:
The Correct Fix would be to make postfix send the username from cert to dovecot instead.

BryanJacobs:
http://postfix.1071664.n5.nabble.com/TLS-client-certificates-and-auth-external-td99246i40.html

The Postfix developers say that the MUA should be changed to not send the AUTH command.

If Dovecot developers say Postfix should change, Postfix developers say the MUAs should change, and the MUAs say that they don't care who changes so long as it's not them (which both Thunderbird and K-9 mail do), what you end up with is nominally open-source software which still can't be made to work right for the users without patching.

Do you have some kind of objection to this patch specifically, or is it just a general feeling that Dovecot is "doing the right thing" so it shouldn't be altered? Is there some way you can help me break the impasse here?

This has been broken for literally years at this point.

cmouse:
At this point I think it would make more sense to open discussion by sending mail to dovecot@dovecot.org.

cmouse:
I don't like this patch as it's introducting a "default username".

BryanJacobs:
Would it be possible for you to propose a type of change to Dovecot that would fix the issue above?

I didn't see an issue with setting a "default username", but if you don't like that idea... what means would be acceptable?

sirainen:
At least it's annoying that this "default username" doesn't make sense except with EXTERNAL mechanism. So auth_external_username setting name would be better. But then .. this is only used in this special case when Postfix talks to dovecot-auth and not in any other situation. So it's a rather special case that would mostly confuse people who aren't using this exact setup. I guess we could think of providing this setting to allow EXTERNAL to always succeed with non-TLS connections as well, e.g. with imap and pop3 as well. So in a way it would be similar to ANONYMOUS authentication. Which makes me wonder: If you enabled ANONYMOUS instead of EXTERNAL, would those clients use it?

sirainen:
Also we're eventually planning on supporting Lua scripting for the whole authentication flow handling. That would allow doing pretty much anything.

BryanJacobs:
The setting is already "auth_external_default_username". It applies to auth EXTERNAL, so it begins with "auth_external". It's overridden by any provided username, so it is the "default".

No, I using ANONYMOUS inside TLS does not work. The clients issue an "AUTH EXTERNAL" command to Postfix.

This is useful for more than Postfix - it deals with any situation where an AUTH EXTERNAL is sent to Dovecot with an empty username. The RFCs do not say the username is mandatory, so technically Dovecot is non-compliant at the moment by rejecting all such requests.

Look, I'll make whatever change you think is right to make this work, you just have to tell me what type of change you're willing to accept. You want a boolean flag that makes Dovecot ignore null usernames and proceed? No problem, just be aware it'll touch more of the code than this. You want something that applies the cert name earlier in the process? Sure, also a more invasive change, but I can do that.  Work with me here, don't make me guess and then have a bunch of PRs rejected or ignored.

Lua scripting, depending on how it was implemented, would likely be harder than this because I would (likely) need to reimplement the code delegating to auth_plain which is already present in the _continue method for external, and is desirable as it calls the existing userdb/passdb setup.

sirainen:
Mainly we don't want to keep adding settings unless there's a really good reason. The more settings, the more testing required and the easier to cause bugs. There are already way too many settings in Dovecot which are causing bugs when implementing changes. So at least when adding a new setting it should be well thought out and how it affects any other potential future design decisions.

Thinking about the ideal Dovecot design related to this:
 * EXTERNAL handling shouldn't trust login processes to verify the client TLS certificate. The current behavior pretty much breaks the whole security design of login-auth processes. We've thought about some day for login processes to send the client TLS certificate to auth process, which would in turn verify it (or maybe there would be some other intermediate tls-worker process doing it.)
 * So the above means that the current behavior of auth protocol using "cert-username" field is only a temporary kludge. Even if Postfix was changed to send it, it could change some day when EXTERNAL handling was made more secure.
 * It might be useful to support other types of EXTERNAL authentication than TLS certificates some day. For example the username could be determined based on the client IP address. Or based on UNIX socket authentication. These would require some kind of new plugins though, or maybe eventually with Lua scripting.
 * Based on these, the username would always be coming from some external source.. I just can't see a default username setting making any sense, except as a kludge to work around client issues.

You mentioned earlier that Dovecot submission service couldn't be used, because Thunderbird doesn't send any AUTH command then. It might make more sense to implement some kind of auto-authentication for Dovecot submission-login. So that if it sees MAIL FROM command, it would attempt to perform AUTH EXTERNAL automatically. @stephanbosch Any thoughts on this idea?

Also, are you sure this can't be made to work if you just forget about EXTERNAL SASL mechanism entirely and use only regular username/password authentication? You could just make Dovecot ignore the password. I always thought this was how the TLS-authentication would get used, since I didn't expect any clients to implement EXTERNAL mechanism.

BryanJacobs:
I definitely can't make non-EXTERNAL mechanisms work - K-9 mail will only present a client certificate if it is set to use EXTERNAL, and in that mode it will not use any other SASL mechanism (and will fail if the EXTERNAL auth command is rejected).

I could fix up the submission service to auth-auth on MAIL FROM - that sounds reasonable. I'm not as big on the idea of entirely reworking how Dovecot deals with SASL EXTERNAL from scratch to be a layer-crossing TLS-validating thing with its own sub-plugins: that sounds possible, but is an absolutely monumental amount of work for a first commit.

If you're willing to accept a PR that makes Submission accept authless mail from within a TLS wrapper (and likely improves logging - I was not impressed with how what came out on max verbose) I am willing to write that.

sirainen:
Tracking internally in DOP-2234. Stephan already wrote some initial code that does implicit EXTERNAL auth on MAIL FROM. Should likely become optional with submission_client_workarounds=implicit-auth-external.

BryanJacobs:
I'm glad to hear that!

When that's ready I'll give a try to routing outbound SMTP via the Dovecot Submission service. Thanks!

BryanJacobs:
Hey!

It's been around three months. I've updated this PR to remain conflict-free with the latest master.

I'm still quite interested in _not_ running a patched Dovecot. I'm still willing to do the engineering to make that happen.

I can't, however, see your "DOP-2234" ticket - I can't find the Jira instance or whathaveyou for it. So I don't know what you're up to internally with respect to making AUTH EXTERNAL work properly with other clients.

Is there something I can do to help here?

cmouse:
Sorry, our process is pretty slow here ... We'll try to get there.

BryanJacobs:
Okay, thanks. I'll check back in another few months.

stephanbosch:
You can play around with it though: https://github.com/stephanbosch/dovecot-core/commits/submission-external-workaround

BryanJacobs:
Looks good. Unrelated to your patch, recent changes have broken compiling Dovecot for me (tests for the new imap hibernation feature rely on being able to open sockets whose names exceed the max permissible length). I'll take the changes for a spin in a week or so when I get some free time to sort it out.

BryanJacobs:
I've finally gotten around to testing out your patched branch. The changes work fine against Dovecot 2.3.14. I had no issues enabling `submission_client_workarounds = implicit-auth-external` and sending email through submission, using TLS certs only, from both K-9 mail and Thunderbird.

Thanks very much! I hope you merge changes like that into Dovecot's mainline.

cmouse:
Implemented via 17efe7a4041358535ede7d78ecc139ea57ece50c

PR #151, opened 2021-01-14
fuzz: fixes targets without ioloop

catenacyber:
For oss-fuzz to work cf https://github.com/google/oss-fuzz/pull/4960


cmouse:
Merged as 709727324b178bdc89c4719b9488a298e29cef38

PR #152, opened 2021-01-15
Fix a corruption of the envelope sender when running sieve scripts

fbellet:
Hi!

I noticed that my inbox folders in mbox format are corrupted, with garbled "From " headers added by dovecot-lda.  Sometimes, the enveloppe sender is replaced by non-printable characters. possibly containing newline, which then generate a visible error in the maillog file, like this one:

```
Jan 15 08:29:31 vm1 dovecot[349261]: lda(testuser)<349261><z6dfMlVEAWBNVAUAkSfQGw>: Error: Next message unexpectedly corrupted in mbox file /var/mail/testuser at 8761229
```
Most of the time, the garbled envelope sender remains unnoticed by the user, but running an ``egrep '^From .*[^[:print:]]' /var/spool/mail/*`` gives a good hint of the presence of the problem.

This is specific to the presence of sieve scripts in the user's home directory. I can reliably reproduce the problem, with dovecot 2.3.13 on a fedora 33 box.

After some debugging, I noticed that function ``sieve_message_substitute()`` obtains the envelope sender by calling ``smtp_address_encode(sender)``, and puts this newly allocated string pointer in ``((struct raw_mailbox *)box)->envelope_sender``.

The problem is that is pointer stops to exist very quickly, when the stack is poped in ``sieve_interpreter_operation_execute()``.

But this string is still required in later sieve scripts operations, especially in ``act_store_execute()`` line 654, with ``((struct raw_mailbox *)mail->box)->envelope_sender``

I'm not sure if duplicating the strings there, and attaching it to the mailbox memory pool is the right way to fix it, but it works for me.

### DIFF
--- a/src/lib-storage/index/raw/raw-storage.c
+++ b/src/lib-storage/index/raw/raw-storage.c
@@ -77,7 +77,7 @@ raw_mailbox_alloc_common(struct mail_user *user, struct istream *input,
 
 	i_assert(strcmp(box->storage->name, RAW_STORAGE_NAME) == 0);
 	raw_box = RAW_MAILBOX(box);
-	raw_box->envelope_sender = envelope_sender;
+	raw_box->envelope_sender = p_strdup(box->pool, envelope_sender);
 	raw_box->mtime = received_time;
 	return 0;
 }


sirainen:
Thanks, tracking internally in DOP-2221.

sirainen:
Can you also share some way of easily reproducing this? Apparently happens with some Sieve scripts?

fbellet:
That's right, it seems to depend on the size of the user sieve script. I have a system sieve-before script calling spamassassin, and a user roundcubemail generated script:
[spamassassin.sieve.txt](https://github.com/dovecot/core/files/5824120/spamassassin.sieve.txt)
and
[roundcube.sieve.txt](https://github.com/dovecot/core/files/5824123/roundcube.sieve.txt)

fbellet:
Here is a debugging session.
[gdb.txt](https://github.com/dovecot/core/files/5824140/gdb.txt)

dovecot and pigeonhole have been rebuild with CFLAGS='-g' using exactly the same configure options than the one used to build the fedora package : [https://kojipkgs.fedoraproject.org//packages/dovecot/2.3.13/2.fc33/data/logs/x86_64/build.log](url). The fedora packages required for the dependencies have been installed as well (the ``BuildRequires`` lines  in [https://src.fedoraproject.org/rpms/dovecot/blob/master/f/dovecot.spec#_40](url))

sirainen:
This is specifically happening only when vnd.dovecot.filter extension is used. And it seems like if you give -f parameter to dovecot-lda it no longer happens.

fbellet:
> This is specifically happening only when vnd.dovecot.filter extension is used. And it seems like if you give -f parameter to dovecot-lda it no longer happens.

In my case, setting another enveloper sender address with the -f parameter doesn't change the problem, just the address given to ``smtp_address_encode()`` is different.

But, I confirm that the kind of filter  program used also changes whether the bug is visible or not. I replaced 'spamc' by 'cat' and the bug is hidden. The bug is still there, but it is not triggered, because in the place where the envelope sender address string has been initially created, it's now overwritten by a zero-filled area, and the null string is a safe value there in [mail_save_copy_default_metadata()](https://github.com/dovecot/core/blob/7e98d506b0643ed2cb19400c38559906feb5abf4/src/lib-storage/mail-copy.c#L42)

For the problem to be visible, I think the overwritten memory area allocated in ``smtp_address_encode()`` must still contain non-null data when read from ``mail_save_copy_default_metadata()`` in my case, and in other places like there:

```
Breakpoint 17, _sieve_result_implicit_keep (result=0x5555555b4c78, rollback=false) at sieve-result.c:954
(gdb) print ((const struct sieve_message_version *)(result->action_env->msgctx->versions->arr->buffer->data))->mail
$27 = (struct mail *) 0x5555555de698
(gdb) print  ((struct raw_mailbox *)($27)->box)->envelope_sender
$28 = 0x555555562a98 "X*VUUU"
(gdb) c
Continuing.

Breakpoint 14, act_store_execute (aenv=0x5555555b4ca8, tr_context=0x5555555b4e10) at sieve-actions.c:654
(gdb) print ((struct raw_mailbox *)mail->box)->envelope_sender
$29 = 0x555555562a98 "X*VUUU"

```

cmouse:
Merged as 9211e803f27bdc57098d03c278dbca572cd618bd

PR #154, opened 2021-01-23
tiny typo

cybernet:
change to https


sirainen:
Thanks, we'll change this. Although doc.dovecot.org is the better place nowadays for documentation, so changed that as well.

sirainen:
The URL is now updated

PR #155, opened 2021-02-11
Fix sending binary code to FTS plugin

grosjo:
This to avoid sending binary data to the FTS plugin when converting attachment in emails

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -558,8 +558,8 @@ DOVECOT_WANT_CASSANDRA
 DOVECOT_SQL
 
 AM_CONDITIONAL([AUTH_LUA_PLUGIN], [test "x$with_lua_plugin" = "xyes"])
-AM_CONDITIONAL([HAVE_LUA], [test "x$with_luajit" = "xyes" || test "x$with_lua" = "xyes"])
-AS_IF([test "x$with_luajit" = "xyes" || test "x$with_lua" = "xyes"],
+AM_CONDITIONAL([HAVE_LUA], [test "x$with_lua" = "xyes"])
+AS_IF([test "x$with_lua" = "xyes"],
   AS_IF([test "x$with_lua_plugin" = "xyes"],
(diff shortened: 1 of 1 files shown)


src/plugins/fts/fts-build-mail.c
```diff
@@ -422,6 +422,7 @@ static int fts_build_data(struct fts_mail_build_context *ctx,
 		    FTS_BACKEND_FLAG_BUILD_FULL_WORDS) != 0) {
 		return fts_build_full_words(ctx, data, size, last);
 	} else {
+		if((!last) && (ctx->body_parser != NULL)) return 0;
 		return fts_backend_update_build_more(ctx->update_ctx, data, size);
```
sirainen, line 426: I think this is leaving out a lot more than just binary data.

grosjo: testing on production server makes it consistent (attachments are indexed, and others field also as usual).

Do you have a use case of not working situation ?

grosjo: To make it proper and exact, the "fts_body_parser_finish" shall be indeed called first, before threst

sirainen: I think what this change does is that it indexes only the last piece of the body parts, no matter what the body part's contents are. So if it's a large blob of binary garbage, it probably indexes the last few (8?) kilobytes of the garbage. But also if it's a large text block it also indexes only the ending of the text.

I am interested though in what kind of emails cause this binary output? FTS is supposed to be indexing only MIME parts that have text/* content-type. Are some such emails wrongly containing binary in its text parts? Or is there something else going wrong? I've also heard that Tika can output garbage from some attachments.

grosjo: No, this is the indexing of attachment, getting out of fts_decoder option.

the indexer (xapian, squat, etc...) receives twice the attachment : first a binary , then the converted data .

The second part shall be the only one

that second part is sent only in the "finish" function

PR #136, opened 2020-09-04
Make metrics compliant with OpenMetrics' standards

mrueg:
running `promtool check metrics` against the current metric output (2.3.11), it mentions:
```
dovecot_imap_command_count counter metrics should have "_total" suffix
dovecot_imap_command_count non-histogram and non-summary metrics should not have "_count" suffix
dovecot_imap_command_duration_usecs_sum counter metrics should have "_total" suffix
dovecot_imap_command_duration_usecs_sum non-histogram and non-summary metrics should not have "_sum" suffix
dovecot_push_notification_http_finished_count counter metrics should have "_total" suffix
dovecot_push_notification_http_finished_count non-histogram and non-summary metrics should not have "_count" suffix
dovecot_push_notification_http_finished_duration_usecs_sum counter metrics should have "_total" suffix
dovecot_push_notification_http_finished_duration_usecs_sum non-histogram and non-summary metrics should not have "_sum" suffix
dovecot_push_notifications_count counter metrics should have "_total" suffix
dovecot_push_notifications_count non-histogram and non-summary metrics should not have "_count" suffix
dovecot_push_notifications_duration_usecs_sum counter metrics should have "_total" suffix
dovecot_push_notifications_duration_usecs_sum non-histogram and non-summary metrics should not have "_sum" suffix
dovecot_stats_uptime_seconds counter metrics should have "_total" suffix
dovecot_storage_http_gets_count counter metrics should have "_total" suffix
dovecot_storage_http_gets_count non-histogram and non-summary metrics should not have "_count" suffix
dovecot_storage_http_gets_duration_usecs_sum counter metrics should have "_total" suffix
dovecot_storage_http_gets_duration_usecs_sum non-histogram and non-summary metrics should not have "_sum" suffix
```

There was also feedback from a Prometheus' developer in https://github.com/prometheus/docs/pull/1738#issuecomment-683635557

@jeffpc @sirainen

PR #157, opened 2021-02-16
stats: Change histogram threshold unit to seconds

mgjm:
In #136 the unit of durations was changed from microseconds to seconds, but the `le` field in histograms is still in microseconds.

This applies the same division by `1e6F` to the histograms threshold, to convert from microseconds to seconds.


jeffpc:
Thanks. The change looks good. We'll get it through the internal CI, etc. and get it committed.

mgjm:
Merged in 9efdef3.

Thanks for the fast reaction and for dovecot in general

PR #158, opened 2021-02-25
if build with -fpermissive => invalid conversion from const void* to …

jrse:
please have a look, 
if i build my c/c++ plugins against it with -fpermissive set its better to cast it explicitly.


sirainen:
Best is actually just remove the function entirely, since it's deprecated.

sirainen:
Ah, this is master-2.3 branch. We haven't been using/updating that branch for a while now. The function is already gone from master.

PR #159, opened 2021-03-06
fix tests for alpine linux as socket name too long

a16bitsysop:


cmouse:
Sorry for taking so long. Why is this an issue?

a16bitsysop:
The current maintainer in alpine has a 3 letter name, if for example I submit a patch it will fail as my name is more than three characters which pushes it over the pathname limit.

cmouse:
Isn't max pathlen in alpine linux 4096? How long paths are you guys using?

cmouse:
Also to me this looks lot more like alpine linux maintainer problem, than dovecot problem. Fixing it like this will only cause it to break again.

a16bitsysop:
I am not sure of the specific length, I can run it through the CI later to get the error message again.  The test path was very long.

a16bitsysop:
Here is the error again:
https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/489469#L6518

cmouse:
ok, we'll discuss what do to with this.

a16bitsysop:
Thanks

sirainen:
Merged https://github.com/dovecot/core/commit/9a3e0d099044d3a7478c3a24ccb8990181767f7c

PR #160, opened 2021-04-07
Fix fts-solr

NixNotCastey:
Send status.messages instead of status.uidnext as rows param to solr.
status.uidnext can return value above int range which rise error 400 - bad request.
Rows parameter control amount of results returned for query, and there is no need to return more results than messages on mailbox.


sirainen:
Should also change STATUS_UIDNEXT to STATUS_MESSAGES, although practically it doesn't currently make a difference.

NixNotCastey:
Thanks, change made. For code cleanliness sake.

cmouse:
I do wonder if we should just use SOLR_MAX_ROWS always?

sirainen:
SOLR_MAX_ROWS is a bit ugly kludge though. The only correct IMAP behavior is to return all the matches, not truncate them to some arbitrary length. I guess one possibility would be to change SOLR_MAX_ROWS to INT32_MAX and use it everywhere?

cmouse:
DOP-2333

cmouse:
Will be handled internally.

PR #162, opened 2021-05-19
imap: imap-sync - Fix VANISHED response with UID=1

splintermail-dev:
Fixes a bug introduced in `4eb3f6f27`, where a UID EXPUNGE 1 command
resulted in a VANSIHED 0:1 response, which is invalid IMAP syntax
(sequence sets must have nonzero values).

The change of `4eb3f6f27` was:

```diff
-               start_uid = 0; prev_uid = (uint32_t)-1;
+               start_uid = 0; prev_uid = 0;
                for (seq = seqs[i].seq1; seq <= seqs[i].seq2; seq++) {
                        mail_set_seq(ctx->mail, seq);
-                       if (prev_uid + 1 != ctx->mail->uid) {
+                       if (prev_uid != ctx->mail->uid - 1) {
```

Notice the second change is mathematically identical, but the first is not.  The effect is that the `if` clause previously executed every first iteration, but after the change, it would miss that first iteration whenever `ctx->mail->uid` was 1.

The fix is just to explicitly include the `prev_uid == 0` case, which make the intended execution clearer anyway.

### DIFF
--- a/src/imap/imap-sync.c
+++ b/src/imap/imap-sync.c
@@ -430,7 +430,7 @@ static void imap_sync_vanished(struct imap_sync_context *ctx)
 		start_uid = 0; prev_uid = 0;
 		for (seq = seqs[i].seq1; seq <= seqs[i].seq2; seq++) {
 			mail_set_seq(ctx->mail, seq);
-			if (prev_uid != ctx->mail->uid - 1) {
+			if (prev_uid == 0 || prev_uid + 1 != ctx->mail->uid) {
 				if (start_uid != 0) {
 					if (!comma)
 						comma = TRUE;


splintermail-dev:
I also wrote an `expect` script for reproducing the buggy behavior.  It logs in, creates a mailbox, populates it, selects it, and deletes UID 1.  Buggy behavior is observable on stdout.

```tcl
#!/usr/bin/expect

send_user "starting\n"

if { $argc < 3 } {
    send_user "usage: $argv0 OPENSSL_CONNECT_STRING USER PASSWORD\n"
    exit 1
}

spawn openssl s_client -connect [lindex $argv 0]

expect "OK ?CAPABILITY IMAP4rev1"

send "1 LOGIN [lindex $argv 1] [lindex $argv 2]\r"
expect "OK * Logged in"

send "2 ENABLE CONDSTORE QRESYNC\r"
expect "2 OK"

send "3 CREATE TEST_MAILBOX\r"
expect "3 OK"

for { set a 0}  {$a < 10} {incr a} {
    send "A$a APPEND TEST_MAILBOX \{7\}\r"
    expect "+ OK"
    send "hello\r\n\r"
    expect "A$a OK"
}

send "4 SELECT TEST_MAILBOX\r"
expect "4 OK"

send "5 UID STORE 1 FLAGS (\\Deleted)\r"
expect "5 OK"

send "6 UID EXPUNGE 1 FLAGS\r"
expect "6 OK"

send "7 CLOSE\r"
expect "7 OK"

send "8 DELETE TEST_MAILBOX\r"
expect "8 OK"

send "8 LOGOUT\r"
expect "BYE"

exit
```

cmouse:
Hi! Thank you for this, we'll look into it.

sirainen:
Merged in https://github.com/dovecot/core/commit/e903c76ef32ea654d4655b5dc104351db21c249d

PR #163, opened 2021-05-20
New setting in auth_passdb allows to disable the master user for proxied connections

fluxkompensator:
In my case dovecot director needs to connect to an IMAP proxy (perdition) which is not able to interpret the "dovecot master user field" when sent via authentication mechanism imap-login, therefore login fails when the master field is present. As I was not able to get around this setting because there seems to be no possibility to drop the master field when proxying connections I tried to introduce this new setting.

Error message from the remote proxy when the master field is present:
`May 20 05:50:30 testbox dovecot: imap-login: proxy(testuser,10.10.10.1:993): Login failed as user testuser (master masteruser): AUTHENTICATE mechanism not supported, mate: user=<testuser>, method=PLAIN, rip=::1, lip=::1, secured, session=<nfo02LrCbIQAAAAAAAAAAAAAAAAAAAAB>`

The connection here was established to localhost (dovecot director) which has the following passdb setup:
```  
passdb {  
    driver = static  
    args = nopassword=y ssl=any-cert  
    default_fields = proxy=y host=10.10.10.1  
    override_fields =  
    master_proxy = no  
    deny = no  
    master = no  
    pass = no  
    skip = never  
    mechanisms =  
    username_filter =  
    #result_failure = continue  
    #result_internalfail = continue  
    #result_success = return-ok  
  }

### DIFF
--- a/src/auth/auth-request-handler.c
+++ b/src/auth/auth-request-handler.c
@@ -205,7 +205,7 @@ auth_str_append_extra_fields(struct auth_request *request, string_t *dest)
 		}
 		if (fields->master_user != NULL &&
 		    !auth_fields_exists(fields->extra_fields, "master") &&
-            request->passdb->set->master_proxy) {
+		    request->passdb->set->master_proxy) {
 			/* the master username needs to be forwarded */
 			auth_str_add_keyvalue(dest, "master",
 					      fields->master_user);
--- a/src/auth/auth-request-handler.c
+++ b/src/auth/auth-request-handler.c
@@ -204,7 +204,8 @@ auth_str_append_extra_fields(struct auth_request *request, string_t *dest)
 					      request->mech_password);
 		}
 		if (fields->master_user != NULL &&
-		    !auth_fields_exists(fields->extra_fields, "master")) {
+		    !auth_fields_exists(fields->extra_fields, "master") &&
+            request->passdb->set->master_proxy) {
 			/* the master username needs to be forwarded */
 			auth_str_add_keyvalue(dest, "master",
 					      fields->master_user);
--- a/src/auth/auth-settings.c
+++ b/src/auth/auth-settings.c
@@ -124,6 +124,7 @@ static const struct setting_define auth_passdb_setting_defines[] = {
 	DEF(BOOL, deny),
 	DEF(BOOL, pass),
 	DEF(BOOL, master),
+	DEF(BOOL, master_proxy),
 	DEF(ENUM, auth_verbose),
 
 	SETTING_DEFINE_LIST_END
@@ -146,6 +147,7 @@ static const struct auth_passdb_settings auth_passdb_default_settings = {
 	.deny = FALSE,
 	.pass = FALSE,
 	.master = FALSE,
+	.master_proxy = TRUE,
 	.auth_verbose = "default:yes:no"
 };
 
--- a/src/auth/auth-settings.h
+++ b/src/auth/auth-settings.h
@@ -20,6 +20,7 @@ struct auth_passdb_settings {
 	bool deny;
 	bool pass; /* deprecated, use result_success=continue instead */
 	bool master;
+	bool master_proxy;
 	const char *auth_verbose;
 };
 
--- a/src/auth/test-mock.c
+++ b/src/auth/test-mock.c
@@ -88,7 +88,7 @@ void passdb_mock_mod_init(void)
 		.deny = FALSE,
 		.pass = FALSE,
 		.master = FALSE,
-	    .master_proxy = TRUE,
+		.master = TRUE,
 		.auth_verbose = "default"
 	};
 	mock_passdb_mod = passdb_preinit(mock_pool, &set);
--- a/src/auth/test-mock.c
+++ b/src/auth/test-mock.c
@@ -58,6 +58,7 @@ struct auth_passdb_settings mock_passdb_set = {
 	.deny = FALSE,
 	.pass = FALSE,
 	.master = FALSE,
+	.master_proxy = TRUE,
 	.auth_verbose = "default"
 };
 
@@ -87,6 +88,7 @@ void passdb_mock_mod_init(void)
 		.deny = FALSE,
 		.pass = FALSE,
 		.master = FALSE,
+	    .master_proxy = TRUE,
 		.auth_verbose = "default"
 	};
 	mock_passdb_mod = passdb_preinit(mock_pool, &set);

### CODE REVIEW

**Comment by cmouse** [comment]
Hi!

Instead of doing this, you could try configuration:

```
passdb {
  driver = static
  args = nopassword=y proxy=y host=10.10.10.1 ssl=any-cert destuser=%{auth_user}
  master = yes
}
```

**Comment by fluxkompensator** [comment]
Thank you, I tried to use destuser like this in passdb for masteruser:
`
passdb {
  driver = passwd-file
  args = /etc/dovecot/passwd.masteruser nopassword=y proxy=y host=10.10.10.1 ssl=any-cert destuser=%{auth_user}
  master = yes
}
`
also some other combinations.
But still the master user that is entered at imap-login overrides everything when the connection ist proxied (proxy=y).
Even if I try in a passdb afterwards "master:remove" it persists in the proxied connection.
For me it seems the reason for this behavior is here in this part of the code, there is no way out of this if the request gets proxied:

`
	if (!request->auth_only &&
	    auth_fields_exists(fields->extra_fields, "proxy")) {
		/* we're proxying */
		if (!auth_fields_exists(fields->extra_fields, "pass") &&
		    request->mech_password != NULL) {
			/* send back the password that was sent by user
			   (not the password in passdb). */
			auth_str_add_keyvalue(dest, "pass",
					      request->mech_password);
		}
		if (fields->master_user != NULL &&
		    !auth_fields_exists(fields->extra_fields, "master")) {
		    !auth_fields_exists(fields->extra_fields, "master") &&
            request->passdb->set->master_proxy) {
			/* the master username needs to be forwarded */
			auth_str_add_keyvalue(dest, "master",
					      fields->master_user);
		}
	}
`

**Comment by cmouse** [comment]
Hi! I can confirm this looks like a bug in dovecot, however, I think the solution would not be adding new settings to passdb like that, but instead, add some new field like "proxy_no_master" to the fields.

**Comment by sirainen** [comment]
I thought setting "master=" would work.

**Comment by cmouse** [comment]
Then it fails with `May 20 09:32:17 auth: Info: plain(?,127.0.0.1,<YoXJGr3Cjsx/AAAB>): Empty username`

**Comment by cmouse** [comment]
This actually seems to be a bug in how `master=` is handled.

**Comment by cmouse** [comment]
Tracked as DOP-2418

**Comment by sirainen** [comment]
Untested, but maybe master= happens to become fixed with:

```
diff --git a/src/login-common/client-common-auth.c b/src/login-common/client-common-auth.c
index b51ed6c198..cffdc481bc 100644
--- a/src/login-common/client-common-auth.c
+++ b/src/login-common/client-common-auth.c
@@ -206,7 +206,7 @@ static void client_auth_parse_args(struct client *client, bool success,
                else if (strcmp(key, "proxy_not_trusted") == 0)
                        reply_r->proxy_not_trusted = TRUE;
                else if (strcmp(key, "master") == 0)
-                       reply_r->master_user = value;
+                       reply_r->master_user = value[0] == '\0' ? NULL : value;
                else if (strcmp(key, "ssl") == 0) {
                        reply_r->ssl_flags |= PROXY_SSL_FLAG_YES;
                        if (strcmp(value, "any-cert") == 0)
```

**Comment by sirainen** [comment]
Although maybe the fix should be already in auth process so it doesn't even send master.

**Comment by sirainen** [comment]
Oh, and depending on your destination proxy implementation, if you set `master=%{user}` it might already work without patching.

**Comment by fluxkompensator** [comment]
> 
> 
> Oh, and depending on your destination proxy implementation, if you set `master=%{user}` it might already work without patching.

Yes overriding master with a custom field `master=something` works. But only if the destination proxy accepts a master field / authentication mechanism.

For my (maybe special) case I need to delete / empty the field `master=`.
But deleting the existing master setting which gets hard set by `proxy=y` is not possible. Not with `master=` and not with `master:remove` neither works. 
But if you overwrite the master field which gets it's input from the imap-login line when proxied, with `master=something` and after that destroy the field with `master:remove`, this does not remove the field but resets to the one set before by imap-login/proxy=y.

**Comment by sirainen** [comment]
I mean if you set `master` to same as the username, Dovecot would do SASL PLAIN authentication with base64(username\0username\0password). Dovecot handles this as regular non-master authentication since the usernames as the same. So if the destination proxy supports SASL PLAIN, it might also handle this the same way since I remember some real IMAP clients doing this same some years ago at least. Of course, if it doesn't support SASL PLAIN at all then this won't work anyway.

I think the fix will anyway be to make `master=` work. Making `master:remove` work will likely be too difficult, although maybe we can check that also.

**Comment by fluxkompensator** [comment]
Thanks for all the help! 

To investigate this further, I did an tcpdump now to see what's going on, I don't really understand that big difference here, but maybe someone can explain.

When I do a simple login on dovecot director which proxies the login to the remote imap proxy tcpdump looks like this when it is working correctly: (In the telnet session I use the command: `a login testuser testpass`)
```
.#......* OK [CAPABILITY IMAP4rev1 QUOTA] perdition ready on testbox01 0004809e  

.....#..L LOGIN "testuser" "testpass"  

.#......* CAPABILITY IMAP4rev1 LITERAL+ SASL-IR LOGIN-REFERRALS ...  

.#.%...*L OK imap4 85
```

This is fine, but when I use a masteruser there is suddenly an `AUTHENTICATE PLAIN` command involved which breaks the destination proxy as it does not support auth mechanisms: (In the telnet session I use the command: `a login testuser*masteruser masterpass`)
```
."......* OK [CAPABILITY IMAP4rev1 QUOTA] perdition ready on testbox01 0004809e  

....."..L AUTHENTICATE PLAIN  

."......L NO AUTHENTICATE mechanism not supported, mate  
```

Why does this happens when the `master=` field is present and why it doesn't when the field is absent like when not using the masteruser for login?
When I remove the `master=` with the (custom compiled) setting `master_proxy = no` then the `AUTHENTICATE PLAIN` part does not happen at all when using the masteruser for login.

**Comment by fluxkompensator** [comment]
> 
> 
> I mean if you set `master` to same as the username, Dovecot would do SASL PLAIN authentication ...

I tried to set `master=testuser` but the same thing happens dovecot issues a `AUTHENTICATE PLAIN` command on the remote, therefore it fails.

**Comment by fluxkompensator** [comment]
So I found another workaround without modifying the source, so I think this pull request is obsolete... sorry for the hassle. 
It seems like my usage is just a special case that needs special configuration. 

I have removed the master passdb entirely, so director is not aware of the masteruser anymore.
Therefore it treats `testuser*masteruser` as a normal user account. 

So here is my final config that works, it's not an easy config but it works in the way I need it:

```
auth_username_format=%{if;%u;*;*masteruser;%0.-10u;%u} # strip "*masteruser" from auth-username  
auth default {  
  mechanisms = plain login  
  # Userdb settings are not used with proxy but there need to be something.  
  userdb {  
    driver = static  
    args = static uid=5000 gid=5000 home=/dev/null  
  }  
  
  # ip's allowed to user masteruser get in the state authenticated by this passdb  
  passdb {  
    driver = static  
    default_fields = allow_nets=127.0.0.0/8,::1,local  
    args = nopassword=y  
    result_success=continue  
  }  
  
  # this passdb proxies masteruser logins without knowing it's a masteruser therefore no "AUTHENTICATE PLAIN" happens  
  passdb {  
    driver = ldap  
    default_fields = allow_nets=127.0.0.0/8,::1,local  
    args = nopassword=y proxy=y host=10.10.10.1 ssl=any-cert user=%u*masteruser  
    result_failure = continue  
    result_internalfail = continue  
    skip = unauthenticated  
  }  
    
  # all other connections without masteruser end up here and get authenticated by password etc...  
  passdb {  
    driver = ldap  
    args = /etc/dovecot/dovecot-ldap.conf.ext  
    skip = authenticated  
  }  
```

cmouse:
Yeah. Although we did identify bug in how `master=` is handled by Dovecot, which we'll fix.

cmouse:
Fixed with f4cf9d69744ef09943baa77860ac63451971030e

PR #164, opened 2021-06-07
lib-oauth2-jwt: Remove 'nbf < iat' check

s3lph:
Dovecot's OAuth2 authentication with local JWT validation contains a check making sure that the `nbf` (valid not before) timestamp does not lie in the past of the `iat` (issued at) timestamp. Otherwise the token is rejected and authentication fails.

However:
- RFC7519 does not mandate any temporal order between these two fields
- It's not uncommon to predate the `nbf` field to account for clock skew.
- Today I've encountered an auth server that predates the `nbf` field by a few minutes. Consequently, Dovecot rejects all tokens issued by this server.

This PR removes this check, as it does not seem to fulfill any purpose other than making Dovecot reject some valid tokens.

### DIFF
--- a/src/lib-oauth2/oauth2-jwt.c
+++ b/src/lib-oauth2/oauth2-jwt.c
@@ -360,8 +360,7 @@ oauth2_jwt_body_process(const struct oauth2_settings *set, const char *alg,
 	}
 
 	/* ensure token dates are not conflicting */
-	if (nbf < iat ||
-	    exp < iat ||
+	if (exp < iat ||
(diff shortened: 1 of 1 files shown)


cmouse:
Have you checked that the auth server's clock isn't wrong!

cmouse:
Ah. I think it might've been a misunderstanding here when making the code. iat can indeed be before nbf as the token might be issued for future use.

cmouse:
Filed as DOP-2440 - it might take a while to get merged.

PR #165, opened 2021-06-08
Add new oauth2 passdb config option 'assume_jwt'

s3lph:
Dovecot's OAuth2 authentication with local JWT validation contains a check making sure that the JWT header field `typ` is present and is exactly `JWT`. However, despite most implementations setting this field, it is actually optional ([RFC 7519 Section 5.1](https://datatracker.ietf.org/doc/html/rfc7519#section-5.1)).

Even though quite uncommon, today I've encountered an auth server that does not set the `typ` header. Consequently, Dovecot rejects all tokens issued by this server.

This PR proposes a new boolean config option for the oauth2 passdb driver backend config: `assume_jwt`
- If `false` (default), the behavior is unchanged, tokens with a missing `typ` header are rejected.
- If `true`, tokens with a missing `typ` header are assumed to be JWT. Tokens with a `typ` other than `JWT` are still rejected though.

I've chosen this approach since it does not introduce any unexpected default behavior, and still respects the `typ` header if present.  "Unexpected behavior" since there could be edge cases where setting this option to `true` could break something. I'm thinking of opaque token implementations that utilize JOSE headers, and thus may look like JWTs on coarse inspection. So this option should only be set to `true` if no such conflicting tokens are to be expected.

### DIFF
--- a/src/auth/db-oauth2.c
+++ b/src/auth/db-oauth2.c
@@ -85,6 +85,8 @@ struct passdb_oauth2_settings {
 	/* Should we send service and local/remote endpoints as X-Dovecot-Auth headers */
 	bool send_auth_headers;
 	bool use_grant_password;
+	/* JWT tokens: If "typ" is missing from header, assume it is a JWT */
+	bool assume_jwt;
 };
 
 struct db_oauth2 {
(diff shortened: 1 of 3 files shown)


cmouse:
I think it would make more sense to implement a `workarounds` setting instead where various workarounds can be added.

cmouse:
DOP-2441

cmouse:
We'll handle this internally.

cmouse:
Fixed as 3adea3c254322b518d0e271c44da2ab104d1e1c7

PR #166, opened 2021-06-30
Support building without Lua (the default)

illumino:
Building dovecot 2.3.15 failed on src/lib-dict's test-dict-client with:
gcc: error: ./.libs/libdict_lua.a: No such file or directory
As the code was building without Lua support but including a Lua variant of the libdict library unconditionally.

This problem did not occur in 2.3.14, but I have not tested 2.3.14.1
It was introduced with 1e67b93cc9f469bb144fc890a6cb9b9b9cd20811

### DIFF
--- a/src/lib-dict/Makefile.am
+++ b/src/lib-dict/Makefile.am
@@ -1,6 +1,5 @@
 noinst_LTLIBRARIES = \
-	libdict.la \
-	libdict_lua.la
+	libdict.la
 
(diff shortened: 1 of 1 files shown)


cmouse:
Thank you for your pull request, we'll take a look.

cmouse:
Fixed with 321c3f31be6ddfcbf4eebf590e20bb3d314b44a5

PR #168, opened 2021-08-08
auth: lib-oauth2: set http basic auth using client_id/secret for post requests

BeryJu:
According to https://datatracker.ietf.org/doc/html/rfc7662#section-2.1 the request to the introspection endpoint should be POST with an Authorization Header set.

Let me know if there's any formatting changes required  or documentation missing.

### DIFF
--- a/src/lib-oauth2/oauth2-request.c
+++ b/src/lib-oauth2/oauth2-request.c
@@ -233,6 +233,13 @@ oauth2_request_start(const struct oauth2_settings *set,
 					       t_strdup_printf("Bearer %s",
 							       input->token));
 	}
+	if (add_auth_bearer &&
+		http_client_request_get_origin_url(req->req)->user == NULL &&
+		set->introspection_mode == INTROSPECTION_MODE_POST) {
+		http_client_request_set_auth_simple(req->req,
+			req->set->client_id,
+			req->set->client_secret);
+	}
 	http_client_request_set_timeout_msecs(req->req,
 					      req->set->timeout_msecs);
 	http_client_request_submit(req->req);


cmouse:
Thank you for this pull request. It seems however that you did not fully read the spec, 

   To prevent token scanning attacks, the endpoint MUST also require
   some form of authorization to access this endpoint, such as client
   authentication as described in OAuth 2.0 [RFC6749] or a separate
   OAuth 2.0 access token such as the bearer token described in OAuth
   2.0 Bearer Token Usage [RFC6750].  The methods of managing and
   validating these authentication credentials are out of scope of this
   specification.

It should not be using the *same* token for authentication, it should be using the client id & secret, or something else. Maybe the correct fix would be to use client id & secret with basic auth always.

BeryJu:
My bad, that's what I get for writing PRs at such late times 😁. 

I don't want to introduce any breaking features, would you prefer post_auth seeing client_id:client_secret auth or a new setting to configure the auth header?

cmouse:
I think we can send the auth header always.

BeryJu:
This is the point where I should've probably opened an issue instead, I don't have enough experience with C (nor do I know this codebase well enough) to correctly base64-encode the Authorzation Basic header with client.

cmouse:
Luckily http client knows how to do this, `http_client_request_set_auth_simple`.

BeryJu:
I wasn't 100% sure under which conditions `add_auth_bearer` is true and which not so I copied the same condition as for the bearer token.

azmeuk:
I stumbled against the same issue with [canaille](https://gitlab.com/yaal/canaille).

cmouse:
Hi!

I just realized that it would make much more sense to put the credentials on the URL, did you try this?

`introspection_url = https://client_id:client_secret@host/... `

BeryJu:
I think that was tried here https://github.com/goauthentik/authentik/issues/1234#issuecomment-895072490

cmouse:
When lib-oauth2 makes http request it uses http_client_request_url_str. (https://github.com/dovecot/core/blob/master/src/lib-oauth2/oauth2-request.c#L213). This in turn uses http url parser with HTTP_URL_ALLOW_USERINFO_PART (https://github.com/dovecot/core/blob/master/src/lib-http/http-client-request.c#L180), which should allow user:password on URL.

azmeuk:
I guess that for authenticating by passing ID and secret in the URL `https://client_id:client_secret@mydomain.tld/oauth/introspect?token=` it is needed that the client `token_endpoint_auth_method` has value `client_secret_basic` and not `client_secret_post`. However [I thought](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) that clients should only support one method. Am I right so far?

If so, would it mean that dovecot would impose `client_secret_basic` as the only authentication method a client should support?

azmeuk:
I see that some tools like authlib only support POST as HTTP method for introspection endpoints. https://github.com/lepture/authlib/blob/master/authlib/oauth2/rfc7662/introspection.py#L37

cmouse:
With the current master code, you can specify any url, not just introspection, with client_id:client_secret. introspection_url was only example here. You should be able to use it with all the settings that accept url.

azmeuk:
Ok. I think I understand what you mean. My understanding of the spec now is slightly different than the OP request.

So this part of the spec says that there should be *some* authentication, but one can use any kind of authentication. So that can be post parameters, or credentials in the URL. Both are OK.

> To prevent token scanning attacks, the endpoint MUST also require
> some form of authorization to access this endpoint, such as client
> authentication as described in OAuth 2.0 [RFC6749] or a separate
> OAuth 2.0 access token such as the bearer token described in OAuth
> 2.0 Bearer Token Usage [RFC6750]. The methods of managing and
> validating these authentication credentials are out of scope of this
> specification.

However, AFAIU this part says that the other parameters `token` and `token_type_hint` must be in a POST payload:

>    The protected resource calls the introspection endpoint using an HTTP
>   POST [RFC7231] request with parameters sent as
>   "application/x-www-form-urlencoded" data as defined in
>   [W3C.REC-html5-20141028].  The protected resource sends a parameter
>   representing the token along with optional parameters representing
>   additional context that is known by the protected resource to aid the
>   authorization server in its response.

cmouse:
The setting introspection_mode controls how this is done:

You can set it as:
 - auth => GET request with Bearer authentication
 - get => GET request as in ?token=foobar
 - post => POST request with token=foobar in content (optional token_type_hint is not there)
 - local => decode & validate JWT token

I would really appreciate a practical example of what does not work. I'm planning on closing this pull request, due to the fact that you can put the client_id & secret in the URL itself.

azmeuk:
I think there is a misunderstanding around the `introspection` term.

What dovecot calls `introspection_url` is, according to the documentation, *the endpoint used to gather extra fields and other information about the user.* However in the OAuth world, this is often referred as [userinfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo).

What `introspection` means for OAuth is generally [token introspection](https://datatracker.ietf.org/doc/html/rfc7662), that would be the `tokeninfo_url` for Dovecot.

Now that this is more clear I can try to reformulate the issue:

- I do not have issue with `introspection_url`, and `introspection_mode` is very convenient to customize its usage.
- According to [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662#section-2.1) the OAuth introspection endpoint, that Dovecot calls `tokeninfo_url`, needs to use HTTP POST. The authentication against this endpoint can be at the discretion of the user. However Dovecot only handles GET requests for `tokeninfo_url`. This causes some problems for instance with libraries like [authlib](https://github.com/lepture/authlib/blob/master/authlib/oauth2/rfc7662/introspection.py#L18) that strictly applies the spec, and does not support GET request for OAuth introspection endpoints. So any IDP built on authlib will not work with Dovecot.

I can think of several solutions:
- replace the GET request with a POST request,
- implement `tokeninfo_mode` parameter that would work similarly to `introspection_mode` and let the users choose how the request should be made. I think this is the best option.

What do you think?

cmouse:
Would it just work if you leave tokeninfo_url out (it's not mandatory) and just use introspection_url together with introspection_mode=post?

azmeuk:
Yes it works. Thanks!
I did not know `tokeninfo_url` was not mandatory, maybe it worth mentioning it in the documentation?

cmouse:
@BeryJu does the suggested way work for you too?

BeryJu:
@cmouse sorry for the late reply, I didn't get a notification from your question hence I haven't looked here. This would also work fine for me

sirainen:
I'm not sure what the end result is here. Is this useful or not? If yes, please reopen against main.

PR #169, opened 2021-09-10
Fix doveadm director flush doc

bmwiedemann:
builtin help texts already showed it correctly


cmouse:
Merged as 00eba31f193070b37ffd17dc827056c668235c2e

PR #170, opened 2021-10-14
Update stats sample config with actually useful default

grinapo:
It is not clear to me whether it is intentional that you distribute default config with a non-functional stats module. 
These are some basic default metrics for (openmetrics compatible) monitoring. Still I need to print all the old metrics and try to find new equivalents. (Observation: _everyone_ on the net, in tutorials and guides use the _old metrics_ due to missing equivalent config. I haven't found _one_ new config example in a half hour search session. Looks disappointing.)

Also included `address = 127.0.0.1` in the example, either to show its usage or to provide a safer default.

### DIFF
--- a/doc/example-config/conf.d/10-metrics.conf
+++ b/doc/example-config/conf.d/10-metrics.conf
@@ -36,6 +36,57 @@
 #  filter = event=mail_delivery_finished
 #  group_by = duration:exponential:1:5:10
 #}
+#
+#metric push_notifications {
+#  filter = event=push_notification_finished
+#}
+
+#metric auth_client_userdb_lookup {
+# filter = event=auth_client_userdb_lookup_finished
+#}
+#
+#metric auth_client_passdb_lookup {
+# filter = event=auth_client_passdb_lookup_finished
+#}
+#
+#metric auth_master_client_login {
+# filter = event=auth_master_client_login_finished
+#}
+#
+#metric client_connection {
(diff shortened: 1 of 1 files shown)


cmouse:
> > Is this still missing something from the old stats? For operating system metrics such as memory usage, disk et al, I would recommend using node-exporter, and I think the docs need to be updated to reflect this.
> 
> I am not sure whether you really wanted to ask **me**. I wanted to get metrics, went to the docs, it says old metrics are _old_ and new are _new_, so I went and tried to use the new but the neither docs nor example configs contain a working setup for the new one.
> 
> Also I don't know which node exporter you mean, but https://github.com/kumina/dovecot_exporter uses the **old** statistics to get metrics. (And on the mailing list people were sent to https://doc.dovecot.org/configuration_manual/stats/openmetrics/ which suggest to use the new builtin system -- except there is no working setup as of now.)
> 
> Or would you suggest not to use the new stats module at all? I'm puzzled.

An attempt to see if you had an idea... Anyways node-exporter is https://github.com/prometheus/node_exporter which lets you export system metrics like CPU usage, disk usage, memory usage etc, which is more recommended than using old-stats for this.

The new setup should be used, but it's different ideology from the old-stats, which explains why people still use it. We should really ship more metrics with it, and your work helps in that.

cmouse:
Please reopen against main.

PR #171, opened 2021-11-11
replication: Default replication sync time was never set

slusarz:
REPLICATION_SYNC_TIMEOUT_SECS is never used

### DIFF
--- a/src/plugins/replication/replication-plugin.c
+++ b/src/plugins/replication/replication-plugin.c
@@ -348,6 +348,7 @@ static void replication_user_created(struct mail_user *user)
 
 	ruser = p_new(user->pool, struct replication_user, 1);
 	ruser->module_ctx.super = *v;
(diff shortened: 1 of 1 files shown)


cmouse:
Please reopen against main

sirainen:
Why was this change done? It's intentional that sync_timeout isn't set, because it changes the replication to become synchronous. Normally we want to do async replication.

slusarz:
I'm confused.  The commit is a cleanup to actually use REPLICATION_SYNC_TIMEOUT_SECS.  Otherwise, the commit should be to remove that constant, which I'm fine with.

sirainen:
Ah, yes, it should never have been added in the first place.

sirainen:
Looks like there are many such macros -> DOV-5159

slusarz:
In that case, there's no need for this patch and it won't be updated for main.

PR #172, opened 2021-11-24
Update maildir-storage.c

AnneBennett:
Check mailbox access using effective UID instead of real UID.  This change is not yet tested.


AnneBennett:
Support for root-squashed NFS filesystems, part 2: maildir.

sirainen:
Same as https://github.com/dovecot/core/pull/173

PR #174, opened 2021-12-15
fix panic in mbox_sync_headers_add_space on 32bit architectures

dinoex:
with dovecot dovecot-2.3.15 and dovecot-2.3.17,
the imap process dies when switching folders with:

dovecot[14023]: imap(user1)<30453><+AdUn6TSe9jAqAoq>: Panic: file mbox-sync-rewrite.c: line 119 (mbox_sync_headers_add_space): assertion failed: (start_pos < data_size)

Enviroment: FreeBSD 12.2 i386, mbox storage
The mbox file in questions is > 2GB.

I fixed the problem by replacing the 32bit "size_t" with "off_t" in mbox-sync-rewrite.c

Please add the patch below to the next release.

### DIFF
--- a/src/lib-storage/index/mbox/mbox-sync-rewrite.c
+++ b/src/lib-storage/index/mbox/mbox-sync-rewrite.c
@@ -93,7 +93,7 @@ static int mbox_fill_space(struct mbox_sync_context *sync_ctx,
 void mbox_sync_headers_add_space(struct mbox_sync_mail_context *ctx,
 				 size_t size)
 {
-	off_t data_size, pos, start_pos;
+	uoff_t data_size, pos, start_pos;
 	const unsigned char *data;
 	void *p;
 
@@ -105,6 +105,7 @@ void mbox_sync_headers_add_space(struct mbox_sync_mail_context *ctx,
 		/* update the header using the existing offset.
 		   otherwise we might chose wrong header and just decrease
 		   the available space */
+		i_assert(ctx->mail.offset >= ctx->hdr_offset);
 		start_pos = ctx->mail.offset - ctx->hdr_offset;
 	} else {
 		/* Append at the end of X-Keywords header,
@@ -150,7 +151,7 @@ static void mbox_sync_header_remove_space(struct mbox_sync_mail_context *ctx,
 					  size_t start_pos, size_t *size)
 {
 	const unsigned char *data;
-	off_t data_size, pos, last_line_pos;
+	uoff_t data_size, pos, last_line_pos;
 
 	/* find the end of the LWSP */
 	data = str_data(ctx->header);
--- a/src/lib-storage/index/mbox/mbox-sync-rewrite.c
+++ b/src/lib-storage/index/mbox/mbox-sync-rewrite.c
@@ -93,7 +93,7 @@ static int mbox_fill_space(struct mbox_sync_context *sync_ctx,
 void mbox_sync_headers_add_space(struct mbox_sync_mail_context *ctx,
 				 size_t size)
 {
-	size_t data_size, pos, start_pos;
+	off_t data_size, pos, start_pos;
 	const unsigned char *data;
 	void *p;
 
@@ -150,7 +150,7 @@ static void mbox_sync_header_remove_space(struct mbox_sync_mail_context *ctx,
 					  size_t start_pos, size_t *size)
 {
 	const unsigned char *data;
-	size_t data_size, pos, last_line_pos;
+	off_t data_size, pos, last_line_pos;
 
 	/* find the end of the LWSP */
 	data = str_data(ctx->header);


sirainen:
This doesn't seem right.. All of these variables are supposed to be pointing to mail header, so the full mbox file size shouldn't matter. So either I'm misunderstanding something, or this patch happens to somehow accidentally work around the main problem. I'm guessing the problem is already earlier with mail.offset vs hdr_offset being wrong. Can you see if this assert triggers?

```
diff --git a/src/lib-storage/index/mbox/mbox-sync-rewrite.c b/src/lib-storage/index/mbox/mbox-sync-rewrite.c
index eeb4878e90..7a0543d014 100644
--- a/src/lib-storage/index/mbox/mbox-sync-rewrite.c
+++ b/src/lib-storage/index/mbox/mbox-sync-rewrite.c
@@ -105,6 +105,7 @@ void mbox_sync_headers_add_space(struct mbox_sync_mail_context *ctx,
                /* update the header using the existing offset.
                   otherwise we might chose wrong header and just decrease
                   the available space */
+               i_assert(ctx->mail.offset >= ctx->hdr_offset);
                start_pos = ctx->mail.offset - ctx->hdr_offset;
        } else {
                /* Append at the end of X-Keywords header,
```

sirainen:
Oh, right, you're changing start_pos to a signed integer. So most likely it becomes negative which makes it pass the assert check. The behavior is still wrong and it might corrupt the mbox.

dinoex:
Running with the new assert build in adding to my patch causes no issue.
The assert is not triggered.
The mailbox itself was/is still valid.

off_t is signed, but smaller unsigned assigned should be safe on 32bit.
on 64bit the size is the same, so we loose a bit, I have not audited that case.

Building with `-Wconversion` shows more possible problems,
and regressions on 64bit with my patch.

By using uoff_t instead reduces some of the warnings.
on 64bit with uoff_t I see no regression in the warnings.

`doveadm -v force-resync` -u user INBOX.Folder
gives no waning nor error nor panic.

sirainen:
I'd just wish to understand exactly what is broken and why a change is fixing it. Currently I don't see why doing these changes would fix anything. I guess they could, but I don't see why/how.

dinoex:
To clarify, my guess what is happening:

The problem I see is the panic above on a mbox folder > 3GB.
I verified the mbox file is vaild,
I rebuild it with no dovecot tool to make sure it was not corrupted.

The panic seem not to happen on 64bit servers,
where size_t and uoff_r have the same number of bits.

But the panic occurs on 32bit, where the position > 2GB does not fit in size_t and gets truncated somewhere. uoff_t is 64bit and can holds positions > 2GB.

dinoex:
FreeBSD clang version 10.0.1
Target: i386-unknown-freebsd12.2
Building the release with `-Wconversion`
I get `34 warnings generated.`
Building with patches I get `28 warnings generated.`

List of warnings reduced:
```
mbox-sync-rewrite.c:144:38: warning: implicit conversion changes signedness: 'unsigned long long' to 'off_t' (aka 'long long') [-Wsign-conversion]
        ctx->mail.space = (pos - start_pos) + size;
                        ~ ~~~~~~~~~~~~~~~~~~^~~~~~
mbox-sync-rewrite.c:135:29: warning: implicit conversion loses integer precision: 'uoff_t' (aka 'unsigned long long') to 'size_t' (aka 'unsigned int') [-Wshorten-64-to-32]
        mbox_sync_move_buffer(ctx, pos, size, 0);
        ~~~~~~~~~~~~~~~~~~~~~      ^~~
mbox-sync-rewrite.c:137:43: warning: implicit conversion loses integer precision: 'uoff_t' (aka 'unsigned long long') to 'size_t' (aka 'unsigned int') [-Wshorten-64-to-32]
        p = buffer_get_space_unsafe(ctx->header, pos, size);
            ~~~~~~~~~~~~~~~~~~~~~~~              ^~~
mbox-sync-rewrite.c:141:30: warning: implicit conversion loses integer precision: 'uoff_t' (aka 'unsigned long long') to 'size_t' (aka 'unsigned int') [-Wshorten-64-to-32]
                ctx->header_first_change = pos;
                                         ~ ^~~
mbox-sync-rewrite.c:201:31: warning: implicit conversion changes signedness: 'unsigned long long' to 'off_t' (aka 'long long') [-Wsign-conversion]
                ctx->mail.space = data_size - last_line_pos;
                                ~ ~~~~~~~~~~^~~~~~~~~~~~~~~
mbox-sync-rewrite.c:169:30: warning: implicit conversion loses integer precision: 'uoff_t' (aka 'unsigned long long') to 'size_t' (aka 'unsigned int') [-Wshorten-64-to-32]
                        start_pos = last_line_pos = pos+1;
                                  ~ ~~~~~~~~~~~~~~^~~~~~~
mbox-sync-rewrite.c:183:54: warning: implicit conversion loses integer precision: 'unsigned long long' to 'size_t' (aka 'unsigned int') [-Wshorten-64-to-32]
                mbox_sync_move_buffer(ctx, start_pos, 0, data_size - start_pos);
                ~~~~~~~~~~~~~~~~~~~~~                    ~~~~~~~~~~^~~~~~~~~~~

```

dinoex:
FreeBSD clang version 10.0.1
Target: i386-unknown-freebsd12.2
Building the release with` -Wconversion` the compiler reports:

```
mbox-sync-rewrite.c:108:32: warning: implicit conversion loses integer precision: 'unsigned long long' to 'size_t' (aka 'unsigned int') [-Wshorten-64-to-32]
                start_pos = ctx->mail.offctx->hdr_offset;
                          ~ ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
```

sirainen:
Most of these places aren't practically a problem, unless you have a single mail in the mbox file where its header is over 2 GB. And that kind of a mail wouldn't make any sense. So far all of these locations I've seen mentioned shouldn't be causing any problems, because they're dealing with mbox mail header sizes. For example this -Wconversion warning:

```
		start_pos = ctx->mail.offset - ctx->hdr_offset;
```

It's not a problem, because even if ctx->mail.offset is huge, ctx->hdr_offset is supposed to be almost as huge as well. Unless the header size is over 2GB, it's not a problem that start_pos is 32bit.

The types that are used here currently seem to be correct so far that I've seen. So I'd rather not change them to uoff_t because it further confuses what they're supposed to be sizes of.

Can you still reproduce the problem? Best would be if you built with clang and ubsan (`./configure --enable-ubsan`). There it will log on runtime exactly where integer wrapping/truncation happens.

dinoex:
Yes, I can reproduce it. With `--enable-ubsan` maillog shows:

imap: Error: mbox-sync-rewrite.c:108:32: runtime error: unsigned integer overflow: 1792678651 - 6087645300 cannot be represented in type 'unsigned long long'
imap: Error: SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior mbox-sync-rewrite.c:108:32 in

dinoex:
Caveat:
The log may be not related to the panic, cause it is serveral seconds after the panic logged

sirainen:
So this is the situation that I first thought would happen. hdr_offset > mail.offset for some reason, so with the added assert it should have crashed there. The mail.offset=1792678651 seems like it's a proper value. But what is that 6087645300? Did you have 6 GB mbox file? In any case hdr_offset isn't supposed to be larger than mail.offset. So something's already wrong by that point.

dinoex:
Yes another one of the other mbox file has 6089270061 bytes.

dinoex:
another log, the number is way to big.

`imap: Error: mbox-sync.c:1347:24: runtime error: implicit conversion from type 'unsigned long long' of value 18446744073709551531 (64-bit, unsigned) to type 'off_t' (aka 'long long') changed the value to -85 (64-bit, signed)`

dinoex:
With the extra assertion put in place, I get logs from a different source with the same number per mailbox:

`imap: Error: mbox-sync-update.c:59:24: runtime error: implicit conversion from type 'int' of value -597322566 (32-bit, signed) to type 'uoff_t' (aka 'unsigned long long') changed the value to 18446744073112229050 (64-bit, unsigned)`

dinoex:
found this 3 times in the log:

```
imap: Error: mail-index-sync-ext.c:323:22: runtime error: unsigned integer overflow: 2192 + 429496627
2 cannot be represented in type 'unsigned int'
imap: Error: mail-index-sync-ext.c:323:25: runtime error: implicit conversion from type 'ssize_t' (ak
a 'int') of value -1024 (32-bit, signed) to type 'unsigned int' changed the value to 4294966272 (32-
bit, unsigned)
imap: Error: mail-index-sync-ext.c:324:22: runtime error: unsigned integer overflow: 2216 + 429496627
2 cannot be represented in type 'unsigned int'
imap: Error: mail-index-sync-ext.c:324:25: runtime error: implicit conversion from type 'ssize_t' (ak
a 'int') of value -1024 (32-bit, signed) to type 'unsigned int' changed the value to 4294966272 (32-
bit, unsigned)
```

PR #179, opened 2022-02-25
policy server allows empty recipient= in END-OF-MESSAGE state

jnorell:
Fixes a bug in the quota-status which does not return a reply when a message has multiple recipients.

When a message has multiple recipients, a policy server server will receive an empty "recipient=" list in the END-OF-MESSAGE state, which is mishandled in the current quota-status server. (With a single message recipient, the recipient= is specified just as it is with the RCPT state, and that is handled correctly.)

Originally created as #105, see comments there.


cmouse:
fixed with 10afc1ce1c87fe55368f3dbed5ea2e1e231244dc

PR #180, opened 2022-03-01
 OBJECTID IMAP extension

filiphanes:
https://tools.ietf.org/html/rfc8474
THREADID == NIL for now, as it is allowed by RFC


sirainen:
Reopened from https://github.com/dovecot/core/pull/131

Internal Jira: DOP-639 - Looks like we also have a similar conflicting patch by Stephan.

PR #184, opened 2022-05-17
oauth2: Add debug info for active_attribute

smokris:
While configuring OAuth2, I was surprised to find that the `active_attribute` check only happens if the server's response includes that attribute (i.e., I was expecting a missing attribute to mean access-denied, but Dovecot treats it as access-granted).

(My OAuth2 server was including the attribute for users who should have access to Dovecot, and was omitting the attribute for users who should _not_ have access to Dovecot.  Since the attribute was not present in the latter response, Dovecot was skipping the check and letting users access Dovecot's services.  I was able to reconfigure my OAuth2 server to provide the attribute in both cases, so now that problem is resolved.)

This PR attempts to clarify the different cases under which the `active_attribute` check passes, fails, or is skipped, to hopefully assist future users in troubleshooting this.

### DIFF
--- a/src/auth/db-oauth2.c
+++ b/src/auth/db-oauth2.c
@@ -603,12 +603,27 @@ db_oauth2_user_is_enabled(struct db_oauth2_request *req,
 	    *req->db->set.active_value != '\0') {
 		const char *active_value =
 			auth_fields_find(req->fields, req->db->set.active_attribute);
-		if (active_value != NULL &&
-		    strcmp(req->db->set.active_value, active_value) != 0) {
-			*error_r = "Provided token is not valid";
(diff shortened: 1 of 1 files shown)


cmouse:
Thank you for your pull request. The absence of ``active_attribute`` permitting login is a bug in my opinion. Sorry for the long delay, we'll take a look at this.

cmouse:
Hi, this has been fixed in different manner in 70a423b6eb90ab50cdd0b0c9618eae87d9a9433f and a789eb687542e01f1d07af7709203d83cf52518f

smokris:
Sounds good.  Thanks, @cmouse.

PR #185, opened 2022-06-08
Set correct passdb entry in auth_worker_handle_passw

PxPert:
Set the correct passdb associated with the passw request (just like auth_worker_handle_passv already does).
This is needed to fix a problem with this specific configuration:

- auth_cache_verify_password_with_worker=yes
- Multiple passdb entries set, with the first set to "deny = yes"

In this specific case, when an user is allowed to login (and his credentials are cached), the login fails.
This is because the **auth_request_password_verify** method checks for the deny value in the request's passdb entry:

`
	if (request->passdb->set->deny) {
		/* this is a deny database, we don't care about the password */
		return 0;
	}
`

So if the first passdb entry has deny set to true, the above condition would always apply and return 0

### DIFF
--- a/src/auth/auth-worker-server.c
+++ b/src/auth/auth-worker-server.c
@@ -305,6 +305,7 @@ auth_worker_handle_passw(struct auth_worker_command *cmd,
 {
 	struct auth_worker_server *server = cmd->server;
 	struct auth_request *request;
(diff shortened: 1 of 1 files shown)


cmouse:
merged as 3013356de329648b2871a341473b130c1af6fd87

PR #186, opened 2022-07-26
Fix endless loop while iterating with dict-fs

manuelm:
We need to stop iterating if no key can be found.

This bug was introduced in 32e8bd5b5a9d5bc04e5ac6c65ac132ba5c386d64


sirainen:
Thanks. I was wondering how that code came to be like that, but looks like it's a regression caused by 32e8bd5b5a9d5bc04e5ac6c65ac132ba5c386d64 in v2.3.17. I created internal DOP-2925 ticket for this.

cmouse:
merged as 180a06ed49894af4a6dfad9c4959046928dfd117

PR #191, opened 2022-12-07
WIP: Added skip_ssl_verify setting to OX push driver

TitaniumHocker:
This is **WIP** patch, I didn't tested it yet.
Updates about tests will be provided later.
I'll update documentation after this PR will be merged.

### DIFF
--- a/src/plugins/push-notification/push-notification-driver-ox.c
+++ b/src/plugins/push-notification/push-notification-driver-ox.c
@@ -49,6 +49,8 @@ struct push_notification_driver_ox_config {
 
 	char *cached_ox_metadata;
 	time_t cached_ox_metadata_timestamp;
+
+	bool skip_ssl_verify;
 };
 
 /* This is data specific to an OX driver transaction. */
@@ -74,6 +76,8 @@ push_notification_driver_ox_init_global(
 		http_set.request_timeout_msecs = config->http_timeout_msecs;
 		http_set.event_parent = user->event;
 		mail_user_init_ssl_client_settings(user, &ssl_set);
+		if (config->skip_ssl_verify)
+			ssl_set.verify_remote_cert = FALSE;
 		http_set.ssl = &ssl_set;
 
 		ox_global->http_client = http_client_init(&http_set);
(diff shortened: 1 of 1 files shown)


cmouse:
Hi.

You should probably use lua driver instead of modifying this driver, as it is heavily interdependent on Open-Xchange Appsuite.

TitaniumHocker:
> You should probably use lua driver instead of modifying this driver, as it is heavily interdependent on Open-Xchange Appsuite.
In my opinion, using lua scripts is actually adding new point of errors. This is not necessary for just disabling ssl verification.
> Also, if you want to do this, I would recommend adding the certificate (or CA cert) to your system store instead.
In my case it's unavailable option.

sirainen:
We're not much changing v2.3 anymore, and I'm planning on doing larger changes to v2.4 configuration handling which will make this possible in a generic way.

TitaniumHocker:
> We're not much changing v2.3 anymore, and I'm planning on doing larger changes to v2.4 configuration handling which will make this possible in a generic way.

As I see you are **planning** to do changes that will solve my problem.
This patch will solve my problem much earlier then it **may** be solved in one of v2.4.x releases.
It just forwards the setting, nothing complicated.

sirainen:
I mean, unless something unexpected happens, the first v2.4 release will have this solved in a different way (any ssl setting can be specified in any context). I'm actively working on the config rewrite now.

TitaniumHocker:
> I mean, unless something unexpected happens, the first v2.4 release will have this solved in a different way (any ssl setting can be specified in any context). I'm actively working on the config rewrite now.

Could you clarify when the release of v2.4 is expected? The last release was almost half a year ago.

sirainen:
> Could you clarify when the release of v2.4 is expected?

There is no schedule for now. Not anytime soon.

TitaniumHocker:
> 

Based on this information, I can only insist on applying this patch.

cmouse:
2.4 is now released and this can be done much easier in 2.4 config now.

PR #192, opened 2022-12-19
Update legacy /var/run directory to /run

istiak101:
Updated legacy /var/run directory to /run.

`Dec 14 20:15:03 hostname.example.com systemd-tmpfiles[499]: /usr/lib/tmpfiles.d/dovecot.conf:1: Line references path below legacy directory /var/run/, updating /var/run/dovecot → /run/dovecot; please update the tmpfiles.d/ drop-in file accordingly.`

Refs:
https://access.redhat.com/solutions/4154291
https://bugs.launchpad.net/ubuntu/+source/dbus/+bug/1854314

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -185,9 +185,9 @@ AS_HELP_STRING([--with-ssldir=DIR], [SSL base directory for certificates (/etc/s
 AC_SUBST(ssldir)
 
 AC_ARG_WITH(rundir,
-AS_HELP_STRING([--with-rundir=DIR], [Runtime data directory (LOCALSTATEDIR/run/dovecot)]),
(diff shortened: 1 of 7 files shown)


cmouse:
Is /var/run legacy on all kinds of \*NIXes now, including FreeSBD, OpenBSD and Mac? This could be worked round with just using `--rundir=/run/dovecot` with configure.

istiak101:
I think you are right. These are mostly distros that ships with systemd. That leaves BSD systems. In that case all the packages on https://repo.dovecot.org/ would need to be updated.

cmouse:
I'll open internal ticket about that. Thanks!

cmouse:
We'll try to fix this for next CE release.

PR #193, opened 2023-01-02
m4: crypt_xxpg6.m4: Define _DEFAULT_SOURCE for current glibc

fweimer-rh:
Current glibc no longer implements the CRYPT extension, so it does not declare crypt in <unistd.h> in strict standard modes.  The check defines _XOPEN_SOURCE, which enables one of these modes.  Defining _DEFAULT_SOURCE as well again makes available the crypt function prototype.

This avoids a configure check result change with compilers which do not support implicit function declarations.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


cmouse:
Can you provide more information about what kind of thing this fixes?

sirainen:
Merged as dd691d40a30fc1f20f8cab3e4740a668d5f0c56f

PR #195, opened 2023-03-24
Fix violation of strict aliasing.

marxin:
The following issue is already reported here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=997513

and the problem is in the following statement: `*(const uint32_t *)&ptr[(n) * 4]`

that yields a miscompilation when LTO is enabled.

### DIFF
--- a/src/lib/md4.c
+++ b/src/lib/md4.c
@@ -34,23 +34,6 @@
 	(a) = ((a) << (s)) | ((a) >> (32 - (s)))
 
 
-/*
- * SET reads 4 input bytes in little-endian byte order and stores them
- * in a properly aligned word in host byte order.
- *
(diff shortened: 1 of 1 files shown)


cmouse:
Thank you for the pull request, we'll take a look.

cmouse:
Merged as f0c1cf42ea78d22e2674b03fe65f0ee6545c5b99

sirainen:
What about the MD5 code, shouldn't that have also disabled the optimizations?

thesamesam:
@sirainen Good spot, I'll make a PR for that.

thesamesam:
> @sirainen Good spot, I'll make a PR for that.

Done in https://github.com/dovecot/core/pull/195.

PR #196, opened 2023-03-27
lib: md5: Fix strict aliasing violation

thesamesam:
Followup to f0c1cf42ea78d22e2674b03fe65f0ee6545c5b99. It's exactly the same code as in md4, so let's rip it out here too.

Thanks to sirainen for pointing this out.

Bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=997513
Reference: https://github.com/dovecot/core/pull/195

### DIFF
--- a/src/lib/md5.c
+++ b/src/lib/md5.c
@@ -38,20 +38,6 @@
 	(a) = (((a) << (s)) | (((a) & 0xffffffff) >> (32 - (s)))); \
 	(a) += (b);
 
-/*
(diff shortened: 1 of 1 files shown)


thesamesam:
cc @sirainen

cmouse:
Thanks, we'll take a look.

cmouse:
Merged as 6de2e4ac0b9abd4ab09b68f1f7e5f032705080b1

thesamesam:
Thanks!

PR #197, opened 2023-05-03
Fix attachment detection

Roobiz:
Hi there,

## Description
Attachment detection on specific case do not works as expected (inline attachment with encoded filename)

## How to reproduce

Tested on latest dovecot version 2.3.20 with the following option:

`mail_attachment_detection_options = add-flags-on-save`

1. Simply add the following email to a mailbox [sample.eml.txt](https://github.com/dovecot/core/files/11381177/sample.eml.txt)
2. Check out the email flags. The email is flagged with $HasNoAttachment instead of $HasAttachment

## Description of the issue

In some case mime-part could have encoded header parameters. It was originally define in the RFC 5987
https://datatracker.ietf.org/doc/html/rfc5987 for the http headers but some applications use it in mime context.

For example Mail.app, adding an inline attachment named "Capture d’écran 2023-05-03 à 11.11.51.png" will produce this kind of part headers:

```
--Apple-Mail=_7891A530-DCCC-4BEA-8EEA-00884364DDF8
Content-Transfer-Encoding: base64
Content-Disposition: inline;
	filename*=utf-8''Capture%20d%E2%80%99e%CC%81cran%202023%2D05%2D03%20a%CC%80%2011.11.51.png
Content-Type: image/png;
	x-mac-hide-extension=yes;
	x-unix-mode=0644;
	name="=?utf-8?B?Q2FwdHVyZSBk4oCZZcyBY3JhbiAyMDIzLTA1LTAzIGHMgCAxMS4xMS41MS5w?=
 =?utf-8?B?bmc=?="
Content-Id: <2EFC0CEF-22B3-44F5-B767-88063E274334>
```

Attachment detection for inline part defined in the lib-mail/message-part-data was strict on the header parameter and only
check for the parameter==="filename".

## The fix
I've simply add another condition on the inline attachment detection to handle encoded "filename"

### DIFF
--- a/src/lib-mail/Makefile.am
+++ b/src/lib-mail/Makefile.am
@@ -99,6 +99,7 @@ test_programs = \
 	test-message-id \
 	test-message-parser \
 	test-message-part \
+	test-message-part-data \
 	test-message-part-serialize \
 	test-message-search \
 	test-message-size \
@@ -202,6 +203,10 @@ test_message_parser_DEPENDENCIES = $(test_deps)
 test_message_part_SOURCES = test-message-part.c
 test_message_part_LDADD = $(test_libs)
 test_message_part_DEPENDENCIES = $(test_deps)
+
+test_message_part_data_SOURCES = test-message-part-data.c
+test_message_part_data_LDADD = $(test_libs)
+test_message_part_data_DEPENDENCIES = $(test_deps)
 
 test_message_search_SOURCES = test-message-search.c
 test_message_search_LDADD = $(test_libs) ../lib-charset/libcharset.la
--- a/src/lib-mail/message-part-data.c
+++ b/src/lib-mail/message-part-data.c
@@ -588,7 +588,9 @@ bool message_part_is_attachment(struct message_part *part,
 	if (null_strcasecmp(data->content_disposition, "attachment") == 0 ||
 	    (!set->exclude_inlined &&
 	     null_strcasecmp(data->content_disposition, "inline") == 0 &&
-	     message_part_has_parameter(part, "filename", FALSE)))
+	     (message_part_has_parameter(part, "filename", FALSE) ||
+	        message_part_has_parameter(part, "filename*", FALSE)
+	     )))
 		return TRUE;
 	return FALSE;
 }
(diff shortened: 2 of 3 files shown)


sirainen:
Hmm. This seems to be RFC 5987 encoding, but that's intended for HTTP only. I wonder how common this is in emails? There is also RFC 2231, which should use `filename*0` for detection.

Roobiz:
> Hmm. This seems to be RFC 5987 encoding, but that's intended for HTTP only. I wonder how common this is in emails? There is also RFC 2231, which should use `filename*0` for detection.

Yep, this what I've describe but unfortunately some apps use this syntax, at least mail.app on MacOS which is a major mail client.

sirainen:
> Yep, this what I've describe but unfortunately some apps use this syntax, at least mail.app on MacOS, and iPhone Mail on iOS which are a major mail clients.

So it seems.. I guess it could be worth it then. Thanks, I'll add this to our internal Jira.

sirainen:
I'm actually wondering now whether we should fully support decoding the parameter in lib-mail, similar to how rfc2231_parse() works now. It converts the `filename*0` etc parameters into a single `filename`. So should we have rfc5987_parse() which does the same thing, so then attachment detection code wouldn't need to be touched?..

sirainen:
Tracking this internally in DOP-3186

Roobiz:
@sirainen it's up to you to choose the better way to manage this issue. Maybe this could also be a slight update un the existing rfc2231 parse method ? For now my proposal, which I know is not perfect, but its the one with the lowest impact on the codebase on my opinion.
Let me know what solution you want and tell me if you want me to investigate and submit new updates on the codebase to handle this :)

exi:
Any updates on this? I can't search for any mails with attachments because of this.

sirainen:
Actually it looks like the `filename*` is supported by RFC 2231, and our RFC 2231 parser is simply wrong. I guess I read the RFC badly many years ago. We'll fix it in RFC 2231 parser.

cmouse:
Fixed with e128abe74bbe0967e91228b60a889a15a9690e97...35fd288ebbe5a49ddf39c10a82000abd7f4e8c5a

PR #203, opened 2023-07-26
Support a SASL mechanism filter

minfrin:
Adds a new filter to specify configuration by SASL mechanism.

### DIFF
--- a/src/auth/auth-request-fields.c
+++ b/src/auth/auth-request-fields.c
@@ -186,6 +186,9 @@ bool auth_request_import_info(struct auth_request *request,
 	} else if (strcmp(key, "local_name") == 0) {
 		fields->local_name = p_strdup(request->pool, value);
 		event_add_str(event, "local_name", value);
+	} else if (strcmp(key, "mech") == 0) {
+		fields->mech_name = p_strdup(request->pool, value);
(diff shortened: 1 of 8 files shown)


cmouse:
There is already mechanisms filter for passdbs. See https://doc.dovecot.org/configuration_manual/authentication/password_databases_passdb/#passdb-setting

minfrin:
> There is already mechanisms filter for passdbs. See https://doc.dovecot.org/configuration_manual/authentication/password_databases_passdb/#passdb-setting

Alas passdbs isn't enough.

To support external and plain at the same time, plain needs "auth_ssl_username_from_cert = no", while external needs "auth_ssl_username_from_cert = yes".

This doesn't appear to be possible in passdb.

sirainen:
Is that a problem though? With PLAIN authentication the client typically wouldn't present SSL certificate, which makes auth_ssl_require_client_cert irrelevant. And if a valid SSL certificate is presented with PLAIN, it should have the correct username anyway then?

cmouse:
Does not really seem that useful, sorry.

minfrin:
> Does not really seem that useful, sorry.

I've been caught up in other things, will give this another try to make work when I have more time.

PR #204, opened 2023-07-31
Fix SOLR 9 configuration file

enqueue:
This change will make the configuration parseable by SOLR.


cmouse:
Can you explain shortly this change?

enqueue:
The file as it is now in main branch is not well-formed XML. You can quickly see this when opening it in VS Code or a Web Browser.

In line 202, a comment is opened via `<!--` and in line 207 another comment is opened.

I figured that there was some duplicate parts in the file, so I removed the upper half. Feel free to correct this, though.

image

cmouse:
fixed with 28867b1f4f8a79b306d0dbf69d9828964ca7d46e

PR #208, opened 2023-09-15
Avoid usage of PATH_MAX not available on hurd

cgzones:


cmouse:
how about

```
#ifndef PATH_MAX
#define PATH_MAX 4096
#endif
```

in compat.h?

cmouse:
merged as e459141dd2f7513850e5910bcd034a072adad67f

PR #209, opened 2023-09-15
Update solr-config-9.xml

cyclops1982:
I had to make this change to make this work :)


cmouse:
All of that?

cyclops1982:
@cmouse  Not sure why the file changes look so strange, but look at the original file in main, you'll see 
```
root@cypher:~# cat /var/solr/data/dovecot/conf/solrconfig.xml
```
on line 204. I had to delete everything up to that line to get it to work.

cmouse:
Oh fun. How did that ever pass ci

cyclops1982:
I don't think CI does a full run of apache solr :-)

cmouse:
But it does. Have to check..

cmouse:
merged as 3294f9074a1c6f2b64f956a21dea3b8b51b33aaf

PR #210, opened 2023-09-15
Update solr-config-9.xml

cyclops1982:
I had to change this to make it work. looks like a little mistake in the file.


cyclops1982:
Duplicate pull request.

PR #212, opened 2023-09-19
Improve logging of tls by including ssl_security information to log

bleve:
Other commit updates documentation of postfix_address which was not updated when default was changed.


cmouse:
Sorry, this stuff is gonna get changed soon due to other work items, so going to close this.

sirainen:
@cmouse We could perhaps add `tls=<%{ssl_security}>` to default login_log_format_elements? Removal of %c doesn't seem quite right though, since it can also include SSL error texts that can be useful.

PR #214, opened 2023-12-18
lib: test-file-cache - build expected error strings using strerror

Duncaen:
Different libc libraries like musl and glibc have different error strings for the same errno's, to make the tests work as expected with musl libc, use strerror to build the expected error string.


cmouse:
Thank you for your pull request, we'll take it under consideration.

cmouse:
merged as bbc5d4670cffe4c5d5b1887098074182c3a0ec4b

PR #215, opened 2024-01-03
Solr/Tika patches for dovecot fts

johnfawcett:
The pull request contains three patches against dovecot main branch:

- Tika - allow authentication against Tika server. Allows setting username and password in fts_tika configuration
- Solr - place upper bound on number of rows to be returned from solr for single mailbox search
- Solr - skip indexing when message size exceed a configured value. New configuration: fts_max_size. Defaults to existing behaviour (no limit)


cmouse:
We'll look at this well

cmouse:
Looks good, we'll take it up to consideration.

cmouse:
Fixed with https://github.com/dovecot/core/compare/caabbf24793a5e4f82cf4d8a382de3cc5a1fc697%5E...731ca99bef1b4a85ddba83b82ff8922aae578831

PR #216, opened 2024-01-11
storage: remove 500 ms debounce on IMAP IDLE notifications

link2xt:
This commit removes hardcoded 500 ms debounce
from storage that delays all storage notification subscribers such as IDLE and NOTIFY commands.

500 ms debounce constant NOTIFY_DELAY_MSECS
was added in 2009 [1]. Before that Dovecot
was only delivering notifications
when a second-resolution timer
is changed by at least 1, so IDLE notifications
were delayed by half a second on average.

[1] https://github.com/dovecot/core/commit/56fb5d09955b6097f77b341717fd9b70e9f13e7

### DIFF
--- a/src/lib-storage/mail-storage-private.h
+++ b/src/lib-storage/mail-storage-private.h
@@ -455,7 +455,7 @@ struct mailbox {
 	/* Mailbox notification settings: */
 	mailbox_notify_callback_t *notify_callback;
 	void *notify_context;
-	struct timeout *to_notify, *to_notify_delay;
+	struct timeout *to_notify;
 	struct mailbox_notify_file *notify_files;
 
 	/* Increased by one for each new struct mailbox. */
--- a/src/lib-storage/mailbox-watch.c
+++ b/src/lib-storage/mailbox-watch.c
@@ -9,8 +9,6 @@
 #include <fcntl.h>
 #include <sys/stat.h>
 
-#define NOTIFY_DELAY_MSECS 500
-
 struct mailbox_notify_file {
 	struct mailbox_notify_file *next;
 
@@ -19,9 +17,10 @@ struct mailbox_notify_file {
 	struct io *io_notify;
 };
 
-static void notify_delay_callback(struct mailbox *box)
+static void notify_callback(struct mailbox *box)
 {
-	timeout_remove(&box->to_notify_delay);
+	timeout_reset(box->to_notify);
+
 	box->notify_callback(box, box->notify_context);
 }
 
@@ -40,18 +39,7 @@ static void notify_timeout(struct mailbox *box)
 	}
 
 	if (notify)
-		notify_delay_callback(box);
-}
-
-static void notify_callback(struct mailbox *box)
-{
-	timeout_reset(box->to_notify);
-
-	if (box->to_notify_delay == NULL) {
-		box->to_notify_delay =
-			timeout_add_short(NOTIFY_DELAY_MSECS,
-					  notify_delay_callback, box);
-	}
+		notify_callback(box);
 }
 
 void mailbox_watch_add(struct mailbox *box, const char *path)
@@ -97,7 +85,6 @@ void mailbox_watch_remove_all(struct mailbox *box)
 		i_free(file);
 	}
 
-	timeout_remove(&box->to_notify_delay);
 	timeout_remove(&box->to_notify);
 }
 


sirainen:
The worry I have related to performance isn't about IMAP client behavior, but rather Dovecot's mailbox_sync() calls. With no delay IDLEing imap process would call mailbox_sync() every time there is a write to dovecot.index.log, whatever that reason is. If another sync is already running by another process, it would wait a bit for the exclusive lock, but otherwise it would be syncing as rapidly as there are .log file writes. Although syncing isn't necessarily expensive, it's not trivially cheap either. So this might increase the server's CPU usage. I'm not sure if that's more of a theoretical problem, but over the years I've grown to be wary of performance regressions, since they sometimes pop up rather unexpectedly.

missytake:
If anyone wants to try this out, we are hosting this PR as a .deb build on https://build.opensuse.org/package/show/home:deltachat/dovecot so it is easy to install

nimbius:
@sirainen  with respect to the CPU concern in 2009 I believe this would be cause to reject the merge; cost is a real factor at scale.

In 2025 I dont believe its a concern.  The smallest mailserver is still equipped with 4-8 CPU cores.  Scalable IMAP servers are all wildly more performant than this however.  Epyc or Intel, I think the concern is moot

dkg:
I've asked for this change [in](https://salsa.debian.org/debian/dovecot/-/merge_requests/44) [debian's experimental repository](https://bugs.debian.org/1098504), and the maintainer there is looking for more feedback from upstream on this change.  It would be great to understand if this patch is under consideration.

Thanks for dovecot, it has been a dependable workhorse for me for years!

sirainen:
How about something like this, which starts with 0 delay but increases it for a while and periodically resets it back to 0? Could still be tweaked a bit I guess, but I'd think this would work well enough for your use case while guaranteeing that there's no huge flooding of the notify callback:

```
diff --git a/src/lib-storage/mail-storage-private.h b/src/lib-storage/mail-storage-private.h
index 6732bd80e4..3a1dcd4a97 100644
--- a/src/lib-storage/mail-storage-private.h
+++ b/src/lib-storage/mail-storage-private.h
@@ -458,6 +458,13 @@ struct mailbox {
 	struct timeout *to_notify, *to_notify_delay;
 	struct mailbox_notify_file *notify_files;
 
+	/* Next timestamp when notifications can be immediately sent without
+	   delay. */
+	struct timeval notify_delay_reset_time;
+	/* Number of notifications since notify_delay_reset_time was updated.
+	   This is used to calculate an increasing delay. */
+	unsigned int notify_delay_counter;
+
 	/* Increased by one for each new struct mailbox. */
 	unsigned int generation_sequence;
 
diff --git a/src/lib-storage/mailbox-watch.c b/src/lib-storage/mailbox-watch.c
index 659cab3810..bad8bbb5f7 100644
--- a/src/lib-storage/mailbox-watch.c
+++ b/src/lib-storage/mailbox-watch.c
@@ -2,6 +2,7 @@
 
 #include "lib.h"
 #include "ioloop.h"
+#include "time-util.h"
 #include "mail-storage-private.h"
 #include "mailbox-watch.h"
 
@@ -9,7 +10,7 @@
 #include <fcntl.h>
 #include <sys/stat.h>
 
-#define NOTIFY_DELAY_MSECS 500
+#define NOTIFY_DELAY_MAX_MSECS 500
 
 struct mailbox_notify_file {
 	struct mailbox_notify_file *next;
@@ -45,13 +46,27 @@ static void notify_timeout(struct mailbox *box)
 
 static void notify_callback(struct mailbox *box)
 {
-	timeout_reset(box->to_notify);
-
-	if (box->to_notify_delay == NULL) {
-		box->to_notify_delay =
-			timeout_add_short(NOTIFY_DELAY_MSECS,
-					  notify_delay_callback, box);
+	unsigned int timeout_msecs = 0;
+
+	if (box->to_notify_delay != NULL)
+		return;
+
+	/* The first notification is immediate. The next notification is
+	   delayed by 10 ms, then 20 ms, 40 ms, .. up to NOTIFY_DELAY_MAX_MSECS.
+	   After NOTIFY_DELAY_MAX_MSECS has passed since the initial delay,
+	   the delay timer is reset. */
+	if (timeval_cmp(&ioloop_timeval, &box->notify_delay_reset_time) < 0) {
+		timeout_msecs = 10U << box->notify_delay_counter++;
+		if (timeout_msecs > NOTIFY_DELAY_MAX_MSECS)
+			timeout_msecs = NOTIFY_DELAY_MAX_MSECS;
+	} else {
+		box->notify_delay_counter = 0;
+		box->notify_delay_reset_time = ioloop_timeval;
+		timeval_add_msecs(&box->notify_delay_reset_time,
+				  NOTIFY_DELAY_MAX_MSECS);
 	}
+	box->to_notify_delay =
+		timeout_add_short(timeout_msecs, notify_delay_callback, box);
 }
 
 void mailbox_watch_add(struct mailbox *box, const char *path)

```

maikelthedev:
Hi guys, any thoughts on merging this PR ? It could be great if this was part of the latest Dovecot version. 

I have a minor suggestion though, why not make it configurable, a switch in dovecot.conf files. 

```ini
disable_debounce = no
```

The default (or omitted) is "no" which keeps things as they are, while the "yes" applies this patch. Wouldn't that be a reasonable equilibrium where everybody is happy? 🥳

PR #217, opened 2024-01-12
Change of severity level

s-wust:
When this event was introduced, it was logged as warning as it is non-critical for dovecot's stability
It was (maybe by accident...) elevated to error when the i_info()/i_warning()/i_error() calls where replaced with their e_*() equivalents
(see https://github.com/dovecot/core/commit/66ddc39ad78a7dbb1af3352307b0e54a9c9dff96)
To prevent flooding error logs with non-critical errors, it should be reverted to warning


sirainen:
Hmm. Maybe .. but why are you getting those errors/warnings anyway? It's not really expected unless one of the login processes crashes or something. Are you maybe using SMTP or some other external auth that connects to Dovecot?

s-wust:
I can't really figure out. I've seen nothing suspicious in my logs.
And yes, I'm using Postfix SMTP AUTH through Dovecot.

PR #218, opened 2024-03-06
push-notification: fix typo 'module' to '-module'

orbea:
There is a typo in the Makefile.am where it has `module` instead of `-module` which causes a build failure with slibtool.
```
ld: cannot find module: No such file or directory
```
With GNU libtool the typo is silently ignored.

This was reported for Gentoo. https://bugs.gentoo.org/913694


cmouse:
Thank you for the pull request, we'll look into this.

PR #221, opened 2024-07-07
hello i found a bugge, you cant access the passwd files as root

curddd:
because he has uid 0


sirainen:
root is intentionally not allowed to access mails. Even if you change passwd-file code, it fails later on.

PR #223, opened 2024-09-25
replicator: fixed infinity loop on sync replication

catap:
Before this fix, replicator adds the same lookup into callbacks over and over, until it reaches out of memory, and crashes.

This regression was introduced 447e086422f1ab7cc16833583ed70a4af7a84bc5 with initial implementation of replicator.

### DIFF
--- a/src/replication/replicator/replicator-queue.c
+++ b/src/replication/replicator/replicator-queue.c
@@ -333,6 +333,7 @@ replicator_queue_handle_sync_lookups(struct replicator_queue *queue,
 		} else {
 			array_push_back(&callbacks, &lookups[i]);
 			array_delete(&queue->sync_lookups, i, 1);
+			count--;
 		}
 	}
 


cmouse:
Thank you for the merge request, we'll take ir into consideration.

catap:
@cmouse thanks for the very quick response!

I understand that replicator was quite unstable.

Frankly speaking, my thread where I debug this issue for about 10 months is good explanation why you removed it from 2.4.

But, I have cluster and will to make it stable... may I ask you to consider to put replicator back in 2.4? Or at least restore some API that allows me to build it as a dedicated plugin.

As you can see, I'm quite motivated to make it stable enough to work :)

cmouse:
Yeah. The only problem is that there is no replicator in 2.4.

catap:
@cmouse because you had removed it at https://github.com/dovecot/core/commit/4c04e4c30fd4817a8b0e11d04d9681173f696f41 :)

Maybe recover it when it finally works? For some setups, this feature can be a deal breaker.

catap:
@cmouse this code contains a condition:
```c
		if (lookups[i].wait_for_next_push) {
			/* another sync request came while user was being
			   replicated */
			i_assert(user->priority == REPLICATION_PRIORITY_SYNC);
			lookups[i].wait_for_next_push = FALSE;
		}
```
if I understand the replicator code base right, when timeout is happened it may switch the `priority` into something else.

So, this abort will be triggered. Probably the right code is something like:
```c
		if (lookups[i].wait_for_next_push) {
			/* another sync request came while user was being
			   replicated */
			if (user->priority != REPLICATION_PRIORITY_SYNC)
				continue;
			lookups[i].wait_for_next_push = FALSE;
		}
```

catap:
BTW I deployed a version with changes from https://github.com/dovecot/core/pull/223#issuecomment-2373715332 and I let you posted how it behaves

catap:
Well, it survives network issue without any crash:
``
Sep 25 13:42:20 mx1 dovecot: lmtp(kirill@korins.ky)<74439><nFEoIhL382bHIgEAeeMcAQ>: Warning: replication(kirill@korins.ky): Sync failure: Timeout in 10 secs
```

catap:
@cmouse after some aditional thought I had simplified this changes to fixing a typo: missed `count--`. If we delete element, we need to iterate for one element less over array.

catap:
@cmouse I continue track an issue where order UID inside vurtual folders are messed up.

Probably the cause of it is... What virtual forlder should be excluded from synchronisation, because otherwise synchronisation fails with hardcoded error: `Can't create virtual mailboxes` (see: `virtual_mailbox_create`).

So, here it seems that I don't understand how should it works. With `-x virtual.` synchronisation works, but it will mess UID inside virtual folders, and without it... it simple fails.

Have I missed something?

catap:
After reading https://dovecot.org/mailman3/archives/list/dovecot@dovecot.org/message/5LYQGL2ZOMDDUFUOJU2TUBLLQUWEPZSR/ and near here.... I understand that replicator won't be recoverd.

@cmouse anyway, may I kindly ask you to merge this small typo? With hope that it will be included into the next release of 2.3 branch.

cmouse:
We will consider it, there is no plan for next 2.3 release at the moment.

catap:
@cmouse thus, if you plan to release 2.3, may I ask you to consider to include this one as well?

cmouse:
Fixed with d3a097075afb7b5d8955978ed4a2ae99add11f88

PR #224, opened 2024-09-25
Support building on OpenBSD

catap:
Here some micro polish which allows to build the main branch on OpenBSD 7.6 beta.

I haven't tested it a lot, but it defently compiled :)


src/lib-dcrypt/dcrypt-openssl1.c
```diff
@@ -3657,14 +3657,14 @@ dcrypt_openssl_private_key_id_old(struct dcrypt_private_key *key,
...
-	char *pub_pt_hex = NULL;
+	const char *pub_pt_hex = NULL;
 
 #ifdef HAVE_X25519
 	if (IS_XD_CURVE(nid)) {
 		unsigned char buf[128];
 		size_t len = sizeof(buf);
 		EVP_PKEY_get_raw_public_key(priv, buf, &len);
-		pub_pt_hex = OPENSSL_buf2hexstr(buf, len);
+		pub_pt_hex = binary_to_hex_ucase(buf, len);
```
sirainen, line 3667: here too

src/lib-ssl-iostream/iostream-openssl-context.c
```diff
@@ -226,7 +226,9 @@ static int ssl_ctx_use_certificate_chain(SSL_CTX *ctx, const char *cert)
 		ret = 0;
 
 	if (ret != 0) {
+#ifdef HAVE_SSL_CTX_set_current_cert
 		SSL_CTX_select_current_cert(ctx, x);
+#endif
```
sirainen, line 231: I wonder if something breaks now because these aren't called, @cmouse ?

sirainen (changes requested):

src/lib-ssl-iostream/iostream-openssl-context.c
cmouse: I'm not sure, would have to test this. @catap would be nice if you could test out that:

local_name cert {
}

blocks work still correctly with SNI

and that e.g. drwetter's SSL checker gives clean bill of health .

src/lib-dcrypt/dcrypt-openssl1.c
```diff
@@ -3617,14 +3617,14 @@ dcrypt_openssl_public_key_id_old(struct dcrypt_public_key *key,
...
-	char *pub_pt_hex = NULL;
+	const char *pub_pt_hex = NULL;
 
 #ifdef HAVE_X25519
 	if (IS_XD_CURVE(nid)) {
 		unsigned char buf[128];
 		size_t len = sizeof(buf);
 		EVP_PKEY_get_raw_public_key(pub, buf, &len);
-		pub_pt_hex = OPENSSL_buf2hexstr(buf, len);
+		pub_pt_hex = binary_to_hex_ucase(buf, len);
```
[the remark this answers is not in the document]
catap: Ooops, it's defently a kind of missed things that someone usually makes at 1am :) Thanks for catching it!

catap:
Well, I've tried to run it and discoevered that dovecot can't start on OpenBSD and probably on non Linux. It doesn't build and install `imap-hibernate` because condition:
```
AM_CONDITIONAL([BUILD_IMAP_HIBERNATE], [test "$notify" != kqueue -a "$notify" != none])
```
but it is included into `src/config/all-settings.c`.

So, when I try to run dovecot, it fails as:
```
Fatal: service(imap-hibernate) access(/home/catap/src/dovecot/local/libexec/dovecot/imap-hibernate) failed: No such file or directory
```

Thus, I had pushed some small polish, but I haven't tested that SNI works because I can't start it :(

cmouse:
you could just disable imap-hibernate? not sure why it even tries to start up though. do you have service imap-hibernate block in your config?

catap:
> you could just disable imap-hibernate? not sure why it even tries to start up though. do you have service imap-hibernate block in your config?

I don’t but it is added via default settings :)

I’m making a way to exclude it, but it is a bit messy and I’m cleaning it right now.

catap:
I just pushed an updated version. It was tested with keys which is generated by command:
```
openssl req -x509 -newkey rsa:4096 -keyout generic.key -out generic.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=generic"
openssl req -x509 -newkey rsa:4096 -keyout hosta.key -out hosta.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=hosta"
openssl req -x509 -newkey rsa:4096 -keyout hostb.key -out hostb.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=hostb"
```

A test was:
```
openssl s_client -crlf -showcerts -connect 127.0.0.1:10993
openssl s_client -crlf -showcerts -connect 127.0.0.1:10993 -servername hosta
openssl s_client -crlf -showcerts -connect 127.0.0.1:10993 -servername hostb 
```
and confirmation that the rigth cerificate is used.

The used config:
```
# 0.0.0-33328+17fb6744c9-dirty (c228f3c277): /home/catap/src/dovecot/local/etc/dovecot/dovecot.conf
# OS: OpenBSD 7.6 amd64  
# Hostname: localhost
dovecot_config_version = 2.4.0
default_internal_group = catap
default_internal_user = catap
default_login_user = catap
dovecot_storage_version = 2.4.0
ssl_cert = </home/catap/src/dovecot/local/etc/dovecot/generic.pem
ssl_key = # hidden, use -P to show it
service auth {
  name = auth
  unix_listener auth-userdb {
    path = auth-userdb
  }
}
service auth-worker {
  name = auth-worker
}
service dict {
  name = dict
  unix_listener dict {
    path = dict
  }
}
service imap {
  name = imap
}
service imap-login {
  name = imap-login
  inet_listener imap {
    name = imap
    port = 10143
  }
  inet_listener imaps {
    name = imaps
    port = 10993
    ssl = yes
  }
}
service lmtp {
  name = lmtp
  unix_listener lmtp {
    path = lmtp
  }
}
service pop3 {
  name = pop3
}
service pop3-login {
  name = pop3-login
  inet_listener pop3 {
    name = pop3
    port = 10110
  }
  inet_listener pop3s {
    name = pop3s
    port = 10995
    ssl = yes
  }
}
service submission {
  name = submission
}
service submission-login {
  name = submission-login
  inet_listener submission {
    name = submission
  }
  inet_listener submissions {
    name = submissions
  }
}
passdb system {
  driver = passwd
  name = system
}
userdb system {
  driver = passwd
  name = system
}
namespace inbox {
  inbox = yes
  name = inbox
  mailbox Drafts {
    name = Drafts
    special_use = \Drafts
  }
  mailbox Junk {
    name = Junk
    special_use = \Junk
  }
  mailbox Trash {
    name = Trash
    special_use = \Trash
  }
  mailbox Sent {
    name = Sent
    special_use = \Sent
  }
  mailbox "Sent Messages" {
    name = Sent Messages
    special_use = \Sent
  }
}
local_name hosta {
  ssl_cert = </home/catap/src/dovecot/local/etc/dovecot/hosta.pem
  ssl_key = # hidden, use -P to show it
}
local_name hostb {
  ssl_cert = </home/catap/src/dovecot/local/etc/dovecot/hostb.pem
  ssl_key = # hidden, use -P to show it
}
```

catap:
I had discovered one more edge case for OpenBSD which already handled well for FreeBSD. I have added one more commit.

configure.ac
```diff
@@ -729,7 +729,7 @@ libdovecot_headers=
 libdovecot_c_files=
 non_libdovecot_headers=
 non_libdovecot_c_files=
-all_files=`find $srcdir/src -name '*.[[ch]]' | grep -v '/src/config/all-settings.c' | grep -v '/src/lib-settings/' | grep -v '/test-' | xargs grep '\\(struct setting_parser_info [[a-z]]\\)\\|\\(struct service_settings [[a-z]]\\)\\|\\(<settings checks>\\)' | sed 's/:.*//' | sort | uniq`
+all_files=`find $srcdir/src -name '*.[[ch]]' | grep -v '/src/config/all-settings.c' | grep -v '/src/lib-settings/' | grep -v '/test-' | xargs egrep '(struct setting_parser_info [[a-z]])|(struct service_settings [[a-z]])|(<settings checks>)' | sed 's/:.*//' | sort | uniq`
```
sirainen, line 739: I remember reading somewhere egrep is going to be deprecated. Should use grep -E instead.

src/lib-dcrypt/dcrypt-openssl1.c
```diff
@@ -3639,7 +3639,12 @@ dcrypt_openssl_public_key_id_old(struct dcrypt_public_key *key,
 	/* digest this */
 	SHA256((const unsigned char*)pub_pt_hex, strlen(pub_pt_hex), buf);
 	buffer_append(result, buf, SHA256_DIGEST_LENGTH);
-	OPENSSL_free(pub_pt_hex);
+
+	/* XD curve uses binary_to_hex_ucase, no OPENSSL_free */
+#ifdef HAVE_X25519
+	if (!IS_XD_CURVE(nid))
+#endif
+		OPENSSL_free(pub_pt_hex);
```
sirainen, line 3647: Isn't this going to complain because pub_pt_hex is const pointer, or does OPENSSL_free() just cast it away? Anyway, perhaps nicer:

```
	const char *pub_pt_hex = NULL;
	const char *pub_pt_hex_free = NULL;
...
		pub_pt_hex = pub_pt_hex_free =
			ec_key_get_pub_point_hex(EVP_PKEY_get0_EC_KEY(pub));
...
OPENSSL_free(pub_pt_hex_free);
```

Would also avoid the secondary ifdef/if checks.

sirainen (changes requested):

sirainen:
The commit titles also could use a bit of clarifying / making them more consistent with how we usually write them, but I could change them also at the end.

configure.ac
catap: Indeed, `egrep` was removed from POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap04.html

catap: After some thinking the cleaner way is using `$GREP`, `$EGREP` and `$SED` and allow configure to decided which should be used on this platform.

catap:
BTW I think that https://github.com/dovecot/core/pull/224/commits/f8ac266e6ce175cefbc0028352cadfb7bc11c50d worth to be backported to 2.3 as well.

cmouse:
hi, can you drop the TLS changes from this? We fixed them in separate change.

configure.ac
cmouse: I found that using $SED seems to break something else.

catap:
@cmouse do you mean 2.3 branch?

cmouse:
No, I mean main branch.

cmouse:
or I guess can just pick the commits that work and push this forward.

configure.ac
cmouse: it seems that For Whatever Reason, it does not correctly collect all dovecot la files in linux.

catap: I may have a look on it in hour or two

catap: @cmouse may I ask you to share an example that doesn't correctly collected on Linux after this changes?

cmouse: I have no idea, we solved it differently in main (please see 6a337ecd25dfc583ff3f16c8f448a2846d0db261) and the $SED change caused, at least for me, that e.g. lib-lua/libdlua.la was missing from lib-dovecot.

catap: Thus, I haven't see any changes in `src/config/all-settings.c` before this commit, and after that on ubuntu 24.04

catap: @cmouse I had removed `$SED` part and seems that it generates everything that it needs on OpenBSD.

src/lib-dcrypt/dcrypt-openssl1.c
```diff
@@ -106,10 +106,20 @@
...
+#  ifdef NID_X448
+#    define IS_XD_CURVE(nid)						\
 	((nid) == NID_X25519 || (nid) == NID_X448)
-# define IS_ED_CURVE(nid) \
+#  else
+#    define IS_XD_CURVE(nid)						\
+	((nid) == NID_X25519)
+#  endif
+#  ifdef NID_ED448
+#    define IS_ED_CURVE(nid)							\
```
sirainen, line 117: \ is in a bit random place. Not soon after define, not aligned with other \ either

sirainen (changes requested):

src/lib-dcrypt/dcrypt-openssl1.c
cmouse: this is also fixed because in OpenBSD there is NID_Ed448.

configure.ac
```diff
@@ -729,19 +729,19 @@ libdovecot_headers=
 libdovecot_c_files=
 non_libdovecot_headers=
 non_libdovecot_c_files=
-all_files=`find $srcdir/src -name '*.[[ch]]' | grep -v '/src/config/all-settings.c' | grep -v '/src/lib-settings/' | grep -v '/test-' | xargs grep -e 'struct setting_parser_info [[a-z]]' -e 'struct service_settings [[a-z]]' -e '<settings checks>' | sed 's/:.*//' | sort | uniq`
+all_files=`find $srcdir/src -name '*.[[ch]]' | $GREP -v '/src/config/all-settings.c' | $GREP -v '/src/lib-settings/' | $GREP -v '/test-' | xargs $EGREP -e 'struct setting_parser_info [[a-z]]' -e 'struct service_settings [[a-z]]' -e '<settings checks>' | sed 's/:.*//' | sort | uniq`
```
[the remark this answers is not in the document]
catap: Thanks to catch it. I had removed this commit and confirmed that it generates `all-settings.c` as it with this changes.

catap:
@cmouse , @sirainen here the force push which addressed remarks.

cmouse:
Did you look at b24ed132668f6360610a40a5504e86d073b329ec, it should address all those lib-dcrypt changes, and works at least in OpenBSD 7.6

cmouse:
fixed with https://github.com/dovecot/core/compare/98220b678b6137d5b0a49fd0a5fb019fb2de64c8...6f6dc7c1c4d08f0f197b758c4ba8919aad640ada

PR #225, opened 2024-09-30
Remove hard coded paths from solr-config-9.xml

samangh:
Dovecot includes a config file for use with Solr 9. However, this file has two problems:

 * it hard codes the path to the Solr installation
 * it hard codes version numbers of ICU analyser packages

This commit removes hard-coded Solr installation path and uses the variable `${solr.install.dir}` instead. It also uses a regex for loading the ICU analyser packages.


cmouse:
Fixed with 0a2de724657b35b2e0b33e70dcb6a4ee0574e263 and 13be30da9edaa5073e14924730ae0c46f6e6bda6

PR #226, opened 2024-10-27
link against icu-uc for icu >= 76

fundawang:
In icu >= 76, `icu-i18n` and `icu-uc` becomes separated libs, so we should link against `icu-uc` explicitly.


cmouse:
Thank you, we'll take a look.

fundawang:
https://gitlab.com/samba-team/samba/-/commit/05807488fd340751ee976c5f8a367013ff94843e

samba has already done this.

cmouse:
Did you try this with older version too?

fundawang:
> Did you try this with older version too?

Yes, I've tested it in icu 74.1.

PR #230, opened 2024-12-28
Backport two strict aliasing fixes

thesamesam:
Fix miscompilation when LTO is enabled.

(cherry picked from commit f0c1cf4)


cmouse:
Sorry, but this is not going into 2.3

thesamesam:
That's a shame, given it was broken on our end for a while as I was expecting the backport, and we had another report of it. Debian had to do the same. But I of course accept the decision. Thanks for the quick reply in any case.

PR #231, opened 2025-01-04
Use i_info() when time jumps forward by <0.2 seconds

andrew-sayers:
Small time jumps due to NTP corrections are common.  Warning the user too frequently trains them to ignore this message, ignoring potentially serious time jumps.

Reduce the level to `info` if the jump is less than 0.2 seconds, so they can opt out of the common messages but still see the problems.

### DIFF
--- a/src/master/main.c
+++ b/src/master/main.c
@@ -49,6 +49,7 @@
 #define FATAL_FILENAME "master-fatal.lastlog"
 #define MASTER_PID_FILE_NAME "master.pid"
 #define SERVICE_TIME_MOVED_BACKWARDS_MAX_THROTTLE_MSECS (60*3*1000)
+#define SERVICE_TIME_MOVED_FORWARDS_MIN_WARNING_MSECS (200*1000)
 
 struct master_delayed_error {
 	enum log_type type;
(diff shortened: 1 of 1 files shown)


sirainen:
Isn't https://github.com/dovecot/core/commit/ac502af78b43a623312e75939af2d75e5c02e5a1 here? Also why <0.2s specifically? It already ignores <0.1s jumps. I don't see a benefit logging 0.1s .. 0.2s with info-log. Might as well not log them at all, i.e. increase:

#define IOLOOP_TIME_MOVED_FORWARDS_MIN_USECS (100000)

andrew-sayers:
I was looking to make the lowest-impact change I could, but silencing the message altogether works for me.

I get an average of about one message per day (sometimes nothing for a week, sometimes several in a day) warning about the clock moving forward by up to about 0.1001 seconds.  I'm running dovecot on an old Raspberry Pi, so that's probably a decent upper-bound.

0.2 seconds seemed like a round number for the patch above, but if we're silencing the message altogether, how about doing `#define IOLOOP_TIME_MOVED_FORWARDS_MIN_USECS (100100)` and bumping it again if anyone else has a problem?

PR #233, opened 2025-01-31
Update hostpid.c

arjendekorte:
Remove last instance of gethostbyname

### DIFF
--- a/src/lib/hostpid.c
+++ b/src/lib/hostpid.c
@@ -50,20 +50,31 @@ void hostpid_deinit(void)
 
 const char *my_hostdomain(void)
 {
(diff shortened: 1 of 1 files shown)


sirainen:
Does it help fix something to get rid of gethostbyname()?

arjendekorte:
Well, gethostbyname is deprecated and not thread safe. Why use that if there is a thread safe alternative?

sirainen:
I'm not necessarily against the change, but to answer the question: Because the old code is smaller, and it's known to work.

cmouse:
Fixed with fd8cf6a6499f9c2821244864d4a507d7da1bb55f

PR #238, opened 2025-03-28
Make solr-config-9.xml compatible with Solr 9.8.1

samangh:
Solr 9.8.1 has one breaking change, which this commit fixes.  In Solr `<lib/>` directives have been made into an "opt-in" feature in must be explicitly enabled if desired. Unless this is done (or a better alternative is used), the dovecot provided config will fail.  See https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-9.html#configuration-2

This PR removes the `<lib/>` directives, and leaves a comment for a use to instead enable the use of Solr modules instead.  This is one of the recommended fixes by Solr.

This comment also removes the `numVersionBuckets` which is obsolete and ignored.

### DIFF
--- a/doc/solr-config-9.xml
+++ b/doc/solr-config-9.xml
@@ -37,53 +37,12 @@
   -->
   <luceneMatchVersion>9.8</luceneMatchVersion>
 
-  <!-- <lib/> directives can be used to instruct Solr to load any Jars
-       identified and use them to resolve any "plugins" specified in
-       your solrconfig.xml or schema.xml (ie: Analyzers, Request
-       Handlers, etc...).
+  <!-- You need to add enable the use of the `analysis-extras` module, by either:
 
-       All directories and paths are resolved relative to the
-       instanceDir.
-
-       Please note that <lib/> directives are processed in the order
-       that they appear in your solrconfig.xml file, and are "stacked"
-       on top of each other when building a ClassLoader - so if you have
-       plugin jars with dependencies on other jars, the "lower level"
-       dependency jars should be loaded first.
-
-       If a "./lib" directory exists in your instanceDir, all files
-       found in it are included as if you had used the following
-       syntax...
-
-              <lib dir="./lib" />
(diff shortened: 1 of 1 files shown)


doc/solr-config-9.xml
```diff
@@ -37,54 +37,13 @@
...
-       dependency jars should be loaded first.
-
-       If a "./lib" directory exists in your instanceDir, all files
-       found in it are included as if you had used the following
-       syntax...
-
-              <lib dir="./lib" />
+       * modifying the "solr.modules" parameter in solr.xml to include "analysis-extras",
+         e.g. set: <str name="modules">${solr.modules:analysis-extras}</str>
+       * adding "-Dsolr.modules=analysis-extras" to the solr start command line.
```
polarathene, line 44: It seems you could also use the equivalent ENV for [Solr Modules](https://solr.apache.org/guide/solr/latest/configuration-guide/solr-modules.html)? If this is the only module to load it would be `SOLR_MODULES=analysis-extras` AFAIK, as the linked docs mentioned if there are any other modules to load it can be done via `,` delimited list.

The given the command line example also aligns with [these config docs](https://solr.apache.org/guide/solr/latest/configuration-guide/property-substitution.html#jvm-system-properties) which demonstrate the equivalent ENV mapping (_the same as the modules docs explain in this case, `SOLR_MODULES=analysis-extras`_).

It might be worth mentioning the ENV alternative which can be more convenient for containers?

polarathene (commented):
I don't use Solr myself, but while reviewing a community contribution to our docs for this issue, I did come across alternative support via ENV that you might want to also document here?

samangh:
@polarathene Thanks for the feedback. I've modified the PR to include setting of the environment variable as another option.

doc/solr-config-9.xml
samangh: Thanks, I've also added a `SOLR_MODULES` example in the comments.

PR #240, opened 2025-03-31
build: do not modify user-provided CFLAGS variable

jengelh:
```
$ ./configure CFLAGS="-U_FORTIFY_SOURCE -g3" && grep ^CFLAGS Makefile && make
...
CFLAGS = -std=gnu11 -U_FORTIFY_SOURCE -g3 -U_FORTIFY_SOURCE -fstack-protector-strong -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -mfunction-return=keep -mindirect-branch=keep -Wall -W -Wmissing-prototypes -Wmissing-declarations -Wpointer-arith -Wchar-subscripts -Wformat=2 -Wbad-function-cast -fno-builtin-strftime -Wstrict-aliasing=2
...
/usr/include/features.h:435:4: warning: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Wcpp]
  435 | #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
```

This is not how `./configure CFLAGS=...` should work. As per https://www.gnu.org/software/automake/manual/html_node/User-Variables.html CFLAGS is reserved for the user, should not be changed (and also should enjoy highest precedence).

AM_CFLAGS is the right variable to use here.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -267,7 +267,7 @@ AC_C_INLINE
 LT_INIT
 
 # SIZE_MAX is missing without this
-CXXFLAGS="$CXXFLAGS -D__STDC_LIMIT_MACROS"
+AM_CXXFLAGS="$AM_CXXFLAGS -D__STDC_LIMIT_MACROS"
 
 # Use the first x.y.z numbers of the version. Also allow x.y numbering.
 AS_IF([echo "$PACKAGE_VERSION" | grep '^[[0-9]]*\.[[0-9]]*\.[[0-9]]'], [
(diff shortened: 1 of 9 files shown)


cmouse:
You could use EXTRA_CFLAGS instead

jengelh:
dovecot.m4 touches EXTRA_CFLAGS too and suffers from losing flags just like CFLAGS during `./configure --enable-ubsan && make EXTRA_CFLAGS=-O1` (loses e.g. sanitizer flags). Also, EXTRA_CFLAGS is nonstandard and not even documented in the source tree.
Please, let's just make CFLAGS behave right. I already wrote the PR.

sirainen:
This change breaks {{./configure --enable-ubsan}} build. Not really sure why.

cmouse:
merged as 25785c90702c0d01f4abf15da546d2bc3bf8e276

PR #242, opened 2025-05-18
push-notification: fix `fts-backend-flatcurve` Time-of-check time-of-use filesystem race condition

odaysec:
https://github.com/dovecot/core/blob/1b8098aa7880ee50f401c8756195f47574f163eb/src/plugins/fts-flatcurve/fts-backend-flatcurve.c#L703-L708

https://github.com/dovecot/core/blob/1b8098aa7880ee50f401c8756195f47574f163eb/src/plugins/fts-flatcurve/fts-backend-flatcurve.c#L719-L719

To fix the TOCTOU race condition, we should avoid relying on the file name after the `stat` check. Instead, we can use a file descriptor obtained via `open` to perform the necessary operations. Specifically:
1. Replace the `stat` call with an `open` call to obtain a file descriptor.
2. Use `fstat` to check the file's properties using the file descriptor.
3. Use `unlink` only if the file descriptor confirms the file's properties match expectations.
4. Close the file descriptor after the operations are complete.

This approach ensures that the file being checked and the file being operated upon are the same, as the file descriptor directly references the file.

---

#### References
The CERT Oracle Secure Coding Standard for C: [FIO01-C. Be careful using functions that use file names for identification ](https://www.securecoding.cert.org/confluence/display/c/FIO01-C.+Be+careful+using+functions+that+use+file+names+for+identification)
[CWE-367](https://cwe.mitre.org/data/definitions/367.html)

### DIFF
--- a/src/plugins/fts-flatcurve/fts-backend-flatcurve.c
+++ b/src/plugins/fts-flatcurve/fts-backend-flatcurve.c
@@ -700,15 +700,25 @@ int fts_backend_flatcurve_delete_dir(const char *path, const char **error_r)
 	struct stat st;
 	enum unlink_directory_flags unlink_flags = UNLINK_DIRECTORY_FLAG_RMDIR;
 
-	if (stat(path, &st) < 0) {
+	int fd = open(path, O_RDONLY);
+	if (fd < 0) {
 		if (errno == ENOENT)
 			return 0;
 		else {
 			*error_r = t_strdup_printf("Deleting fts data failed: "
-				"stat(%s) failed: %m", path);
+				"open(%s) failed: %m", path);
 			return -1;
 		}
 	}
+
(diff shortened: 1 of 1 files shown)


sirainen:
This could make some sense if we had unlink_directory_fd() function, and the fd was passed to it. But as it is now, there is no practical difference to just using stat() call.

Also this directory isn't in any untrusted/shared path. It won't unexpectedly switch between being a file and a directory, so there is no practical race condition. And even if it did happen, the deletion would just fail, so nothing bad would really happen.

slusarz:
And this is a non-critical error/code-path anyway. It only gets run manually (via doveadm command).

Worst case scenario is that and admin runs the command and any error prevents the directory to be deleted via Dovecot.  An admin can still delete the directory manually after investigating the reason for the denial.

PR #246, opened 2025-06-13
global: More fixes for compiling without LDAP

flyn-org:
I discovered that Dovecot 2.4.0 would not build without ldap.h while trying to update OpenWrt's Dovecot package. I found commit 61fe6bbd, but it appeared incomplete after I backported it. I think there are still some conditionals that need to be added to the build system.

This commit adds a few more conditionals. There is still a problem with `all-settings.c` unconditionally including ldap.h, but I don't yet understand `settings-get.pl` enough to fix this.

Additionally, it might be possible to simplify things, both with my changes and the existing build definitions.

Though incomplete, I hoped this patch would help make progress.

### DIFF
--- a/src/auth/Makefile.am
+++ b/src/auth/Makefile.am
@@ -130,8 +130,12 @@ auth_common_sources = \
 	userdb-prefetch.c \
 	userdb-static.c \
 	userdb-sql.c \
+if HAVE_LDAP
 	$(ldap_sources) \
+endif
+if HAVE_LUA
 	$(lua_sources)
+endif
 
 headers = \
 	auth.h \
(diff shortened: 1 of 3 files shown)


cmouse:
Can you please use 2.4.1 for both dovecot core and pigeonhole? Then use 61fe6bbd4a8a64e0823534a57a8db483019c4901 and 4eddec25d21e2bead9e179b407e7e458ffc51910, and you need https://github.com/dovecot/pigeonhole/commit/b37322a8d338bf999bc9774154dce79c4857d622

cmouse:
Also include 431e328b3b035ddb187526cd13bccf29833aed90

PR #247, opened 2025-06-26
Changes for 2.4.1 for i686 compatibility

pghmcfc:
This PR changes a few files (mainly test files) to get the test suite to pass with linux on i686. This reflects some filter values not fitting in an i686 long integer (32 bits), and fixes for a signed 32-bit time_t. The configure script already does all of the necessary checks; it's just a case of using the results properly.


src/lib-imap/test-imap-date.c
```diff
@@ -25,7 +25,7 @@ static void test_imap_date(void)
 		{ "31-Dec-9999", 253402214400LL },
 #endif
 		/* conversions to maximum values */
-#if TIME_T_MAX_BITS == 31
+#if TIME_T_MAX_BITS == 31 || (TIME_T_MAX_BITS == 32 && defined(TIME_T_SIGNED))
```
sirainen, line 28: Hmm. I think these checks are assuming TIME_T_MAX_BITS=31 with signed 32bit time_t. Also m4/gmtime_max.m4 has:

```
      printf "check failed, assuming "
      i_cv_gmtime_max_time_t=31
```

So I'm thinking it might be better to fix the m4 test to output TIME_T_MAX_BITS=31 instead with 32bit signed time_t. Then these other checks wouldn't need to change (except tm_is_too_large() probably).

sirainen (changes requested):

src/lib-imap/test-imap-date.c
sirainen: Also in this commented case `TIME_T_MAX_BITS == 31` would never happen, so I guess that should be simply `#if  #if TIME_T_MAX_BITS == 31 || (TIME_T_MAX_BITS == 32 && defined(TIME_T_SIGNED)`

src/lib-imap/test-imap-date.c
```diff
@@ -71,7 +71,7 @@ static void test_imap_datetime(void)
 		{ "31-Dec-9999 23:59:59 -2359", 253402300799LL + 23*60*60 + 59*60, -23*60 - 59 },
 #endif
 		/* conversions to maximum values */
-#if TIME_T_MAX_BITS == 31
+#if TIME_T_MAX_BITS == 31 || (TIME_T_MAX_BITS == 32 && defined(TIME_T_SIGNED))
```
sirainen, line 74: here also no 31 check

pghmcfc:
Changed the m4 test to make it return 31 bits instead of 32 if time_t was signed, and that made most of the rest of the changes redundant.

cmouse:
See #275

PR #277, opened 2026-04-21
lib: Foundational hardening of allocation, buffer, and array APIs aga…

uwezkhan:
This patch hardens Dovecot's core memory and buffer APIs against integer overflows.
While Dovecot already uses MALLOC_MULTIPLY in some places, this change systematically introduces array-aware primitives (like i_malloc_array and buffer_append_array) and pushes those checks directly into the ARRAY implementation.

### DIFF
--- a/src/lib/imem.c
+++ b/src/lib/imem.c
@@ -9,21 +9,22 @@ void *i_malloc(size_t size)
 	return p_malloc(default_pool, size);
 }
 
-void *i_malloc_array(size_t count, size_t size)
-{
-	return p_malloc(default_pool, MALLOC_MULTIPLY(count, size));
-}
-
 void *i_realloc(void *mem, size_t old_size, size_t new_size)
 {
-	return p_realloc(default_pool, mem, old_size, new_size);
-}
+	i_assert(old_size < SIZE_MAX);
+	i_assert(new_size < SIZE_MAX);
 
-void *i_realloc_array(void *mem, size_t old_count, size_t new_count, size_t size)
-{
-	return p_realloc(default_pool, mem,
-			 MALLOC_MULTIPLY(old_count, size),
-			 MALLOC_MULTIPLY(new_count, size));
+	if (old_size == new_size)
+		return mem;
+
+	if (old_size != 0 && new_size == 0)
+		i_panic("Trying to reallocate %zu -> %zu bytes",
+			old_size, new_size);
+
+	if (mem == NULL)
+		return i_malloc(new_size);
+
+	return p_realloc(default_pool, mem, old_size, new_size);
 }
 
 char *i_strdup(const char *str)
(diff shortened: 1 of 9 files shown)


src/lib/imem.h
```diff
@@ -10,8 +10,11 @@ extern pool_t default_pool;
 	p_realloc_type(default_pool, mem, type, old_count, new_count)
 
 void *i_malloc(size_t size) ATTR_MALLOC ATTR_RETURNS_NONNULL;
+void *i_malloc_array(size_t count, size_t size) ATTR_MALLOC ATTR_RETURNS_NONNULL;
```
sirainen: Thanks for catching these many missing multiplication places! I'm not sure about these `[ip]_malloc_*` functions though. We already have [ip]_new() and [ip]_realloc_type() which are better to use when possible. I see a few places where they should have been used, but aren't. But are there any places were [ip]_malloc_array() would be useful? Seems like it might just encourage people to use it when they should rather be using [ip]_new().

sirainen (changes requested):

uwezkhan:
Thanks, that makes sense. I’ve removed the malloc_array helpers and switched everything to i_new / p_new and p_realloc_type where possible. Let me know if anything else needs tweaking.

src/lib/imem.c
```diff
@@ -11,6 +11,19 @@ void *i_malloc(size_t size)
 
 void *i_realloc(void *mem, size_t old_size, size_t new_size)
 {
+	i_assert(old_size < SIZE_MAX);
+	i_assert(new_size < SIZE_MAX);
```
sirainen, line 15: Already checked by p_malloc() / p_realloc() that size isn't > POOL_MAX_ALLOC_SIZE

src/lib/imem.c
```diff
@@ -11,6 +11,19 @@ void *i_malloc(size_t size)
...
+
+	if (old_size == new_size)
+		return mem;
```
sirainen, line 18: Optimization, but not really necessary..

src/lib/imem.c
```diff
@@ -11,6 +11,19 @@ void *i_malloc(size_t size)
...
+
+	if (old_size != 0 && new_size == 0)
+		i_panic("Trying to reallocate %zu -> %zu bytes",
+			old_size, new_size);
```
sirainen, line 22: new_size==0 panic is already in p_malloc() and p_realloc().

So overall, these i_realloc() seem unnecessary.

Anyway, no need to further update these changes. I did some commit splitting and some other minor changes and we'll get these merged.

uwezkhan:
Thanks! Appreciate the review and the cleanup. 👍

uwezkhan:
Can i get the review on this PR.

sirainen:
It's now queued in our internal review, but at the moment we have a lot of changes waiting for review, so it likely takes a while more still before it gets merged.

sirainen:
Merged as bb2d8814d2e77852953952762ae72c2c31c1dc5a^..6eb492fbdd6c7955890c96c405619538a8fb9223

PR #280, opened 2026-04-30
Add integer overflow checks to p_strarray_dup and new test for large …

netliomax25-code:
p_strarray_dup() accumulates allocation size in a loop:

    size += sizeof(const char *) + strlen(arr[i]) + 1;

This is done without checking for overflow, so `size` can wrap before
the final allocation. Since the pool allocator only validates the final
size, an overflow here can result in allocating a buffer smaller than
intended.

Fix:
Add a SIZE_MAX guard before the addition to prevent overflow during size
accumulation. This follows the pattern used elsewhere (e.g. MALLOC_ADD)
and keeps failure behavior consistent (i_panic on overflow).

Test:
Add a test with a large input array to exercise the accumulation path
and verify behavior is preserved.

### DIFF
--- a/src/lib/strfuncs.c
+++ b/src/lib/strfuncs.c
@@ -962,9 +962,14 @@ const char **p_strarray_dup(pool_t pool, const char *const *arr)
 	char *p;
 	size_t len, size = sizeof(const char *);
 
-	/* @UNSAFE: integer overflow checks are missing */
-	for (i = 0; arr[i] != NULL; i++)
-		size += sizeof(const char *) + strlen(arr[i]) + 1;
+	for (i = 0; arr[i] != NULL; i++) {
+		len = strlen(arr[i]) + 1;
+		/* Check for integer overflow before accumulation */
+		if (len > SIZE_MAX - sizeof(const char *) ||
+		    size > SIZE_MAX - (sizeof(const char *) + len))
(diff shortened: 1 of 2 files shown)


src/lib/strfuncs.c
```diff
@@ -962,9 +962,14 @@ const char **p_strarray_dup(pool_t pool, const char *const *arr)
...
-	/* @UNSAFE: integer overflow checks are missing */
-	for (i = 0; arr[i] != NULL; i++)
-		size += sizeof(const char *) + strlen(arr[i]) + 1;
+	for (i = 0; arr[i] != NULL; i++) {
+		len = strlen(arr[i]) + 1;
+		/* Check for integer overflow before accumulation */
+		if (len > SIZE_MAX - sizeof(const char *) ||
+		    size > SIZE_MAX - (sizeof(const char *) + len))
+			i_panic("p_strarray_dup(): size overflow");
+		size += sizeof(const char *) + len;
```
sirainen, line 971: Hi, thanks, but I've already a bit simpler fix waiting for internal review (we're a bit blocked on code reviews now):

```
	for (i = 0; arr[i] != NULL; i++) {
		size = MALLOC_ADD3(size, sizeof(const char *) + 1,
				   strlen(arr[i]));
	}
```

PR #282, opened 2026-05-02
imap-parser-nesting-limit

uwezkhan:
The IMAP parser currently tracks list_count_limit, but nested list depth is only constrained after lists are closed. This means deeply nested input can still build a large recursive imap_arg tree before the limit is enforced.

Several downstream imap_arg consumers traverse these structures recursively, so excessively deep nesting may lead to stack exhaustion.

This change adds proactive nesting-depth enforcement when opening lists by:

* tracking current nesting depth during parsing
* rejecting input once nesting exceeds the configured list_count_limit
* applying a hard safety ceiling of 100 levels to protect recursive consumers even if a higher limit is configured

The hard ceiling is intentionally conservative and aligns with similar nesting limits elsewhere in the codebase.

Tests are included for both:

* the default hard nesting ceiling
* custom list_count_limit enforcement

### DIFF
--- a/src/lib-imap/imap-parser.c
+++ b/src/lib-imap/imap-parser.c
@@ -20,6 +20,10 @@
 	((c) == '\r' || (c) == '\n')
 
 #define LIST_INIT_COUNT 7
+/* Max. nesting depth for lists. This should be enough for all legitimate
+   uses, but small enough to prevent stack exhaustion in recursive
+   imap_arg consumers. */
+#define IMAP_PARSER_MAX_NESTING_DEPTH 100
 
 enum arg_parse_type {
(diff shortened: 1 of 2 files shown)


cmouse:
fixed with 9a0f8c1066956ea7450ca82b57f904332006a502

uwezkhan:
@cmouse Hi, I saw that the PR was closed after the issue was resolved separately. I had worked on the fix and submitted the PR directly, so I wanted to ask if I would still be eligible for contributor credit or the contributor tag for the effort.

Please let me know. Thank you!

cmouse:
The fix was already made before your contribution. You just forced us to publish it.

PR #283, opened 2026-05-02
lib-imap: enforce nesting depth when opening IMAP lists

uwezkhan:
Recursive consumers of imap_arg may exhaust stack space when processing excessively nested IMAP lists.

This change adds proactive nesting-depth enforcement during parsing so deeply nested list structures are rejected before they are instantiated.

Changes included:

* track active nesting depth in struct imap_parser
* enforce nesting limits in imap_parser_open_list()
* apply list_count_limit during active nesting
* add a hard safety ceiling of 100 levels for recursive consumer protection
* add unit tests covering nesting-limit enforcement

### DIFF
--- a/src/lib-imap/imap-parser.c
+++ b/src/lib-imap/imap-parser.c
@@ -21,6 +21,9 @@
 
 #define LIST_INIT_COUNT 7
 
+/* Maximum depth of nested lists to prevent stack exhaustion in recursive consumers. */
+#define IMAP_PARSER_MAX_NESTING_DEPTH 100
+
 enum arg_parse_type {
 	ARG_PARSE_NONE = 0,
 	ARG_PARSE_ATOM,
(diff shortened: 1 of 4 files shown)


cmouse:
@uwezkhan hi, you inadvertendly published a security issue, can you please next time contact us via our security endpoints for issues that are security related.

uwezkhan:
Apologies for publishing it publicly. I’ll use the security reporting channel for similar findings going forward.

For this case, could you let me know what the recommended next steps are on my side, if any?

cmouse:
Hi, we'll take it from here. This was only a headsup to avoid future accidents.

uwezkhan:
okay

sirainen:
The actual bug is fixed now in https://github.com/dovecot/core/commit/9a0f8c1066956ea7450ca82b57f904332006a502

For the more generic nesting limit, it's a bit more tricky. Most commands don't need nesting, but there are e.g. SEARCH command where valid commands could have a lot of nesting. And it's possible some clients do lots of nesting (at least iOS did some horrible SEARCH queries before). So for now we don't want to limit the nesting everywhere. Need to figure out something at some point how to handle them.

uwezkhan:
@cmouse Hi, I saw that the PR was closed after the issue was resolved separately. I had worked on the fix and submitted the PR directly, so I wanted to ask if I would still be eligible for contributor credit or the contributor tag for the effort.

Please let me know. Thank you!

cmouse:
No, you just forced us to release the fix, which was done independently of your work.

PR #284, opened 2026-05-02
lib: fix nearest_power() overflow handling on 32-bit platforms

uwezkhan:
On 32-bit platforms, `nearest_power()` may return an undersized value
for inputs larger than the maximum representable power of two.

This can cause callers to receive a buffer smaller than requested when
rounding large allocation sizes.

Reject overflow-range inputs explicitly so `nearest_power()` never
returns a value smaller than the requested size.

This change preserves the existing `nearest_power(0) == 1` behavior and
adds regression tests covering overflow-range inputs.


src/lib/test-bits.c
```diff
@@ -74,6 +75,10 @@ static void test_nearest_power(void)
 	test_assert_idx(nearest_power(num-1) == num,    b);
 	test_assert_idx(nearest_power(num  ) == num,    b);
 	/* i_assert()s: test_assert_idx(nearest_power(num+1) == num<<1, b); */
+	/* Test edge case: nearest_power should never return a value smaller than input */
+	num = ((size_t)1 << (sizeof(size_t) * 8 - 1));
```
pghmcfc: Shouldn't the "8" here be "CHAR_BIT"?

src/lib/test-bits.c
```diff
@@ -74,6 +75,10 @@ static void test_nearest_power(void)
...
+	test_assert(nearest_power(num) >= num);
+	test_assert(nearest_power(num + 1) >= num + 1);
```
pghmcfc: I find this last test overflows on i686, testing value 2147483649 (2^31+1) and on x86_64, testing value 9223372036854775809 (2^63+1); shouldn't it be testing that the assertion fails?

uwezkhan: I've updated the tests accordingly

cmouse:
merged as 00120f5ba88fbd04f99ede0720655d03467c2f59

PR #285, opened 2026-05-02
lib: use checked arithmetic in string and path length calculations

uwezkhan:
Several core string and path utility helpers manually compute required
buffer sizes using unchecked length arithmetic such as pos + len + 1.

Replace these calculations with MALLOC_ADD/MALLOC_ADD3 so overflow
is detected explicitly before buffer growth calculations proceed.

This improves robustness of core helper routines and ensures oversized
length summations fail safely rather than relying on downstream
allocator assumptions.

Updated call sites include:

* vstrconcat()
* p_strarray_join_n()
* path normalization / directory resolution helpers


cmouse:
merged as 651436ab5258c88cdb8e3ab21df8dd6422708742

PR #289, opened 2026-05-23
lib-oauth2: jwt - Avoid use-after-realloc of jwt_node front pointer

rootvector2:
oauth2_jwt_copy_fields() captures subroot from array_front(&nodes)
then calls array_append_space(&nodes) inside the inner loop. When the
append cannot extend the buffer in place it relocates it, so subroot
aliases the previous slot and later subroot->prefix / subroot->array
reads operate on the abandoned location. Copy the front node by value
before the append.

### DIFF
--- a/src/lib-oauth2/oauth2-jwt.c
+++ b/src/lib-oauth2/oauth2-jwt.c
@@ -387,6 +387,7 @@ oauth2_jwt_copy_fields(ARRAY_TYPE(oauth2_field) *fields,
 
 			if (!json_node_is_singular(jnode)) {
 				root = array_append_space(&nodes);
+				subroot = array_front(&nodes);
 				root->root = json_tree_node_get_child(tnode);
 				root->array = json_node_is_array(jnode);
 				if (jnode->name == NULL)
--- a/src/lib-oauth2/oauth2-jwt.c
+++ b/src/lib-oauth2/oauth2-jwt.c
@@ -387,6 +387,8 @@ oauth2_jwt_copy_fields(ARRAY_TYPE(oauth2_field) *fields,
 
 			if (!json_node_is_singular(jnode)) {
 				root = array_append_space(&nodes);
+				/* array_append_space() may have reallocated the
+				   buffer, so refresh subroot to the new location. */
 				subroot = array_front(&nodes);
 				root->root = json_tree_node_get_child(tnode);
 				root->array = json_node_is_array(jnode);


src/lib-oauth2/oauth2-jwt.c
```diff
@@ -378,8 +378,11 @@ oauth2_jwt_copy_fields(ARRAY_TYPE(oauth2_field) *fields,
 	root->root = json_tree_get_root_const(tree);
 
 	while (array_count(&nodes) > 0) {
-		const struct jwt_node *subroot = array_front(&nodes);
-		const struct json_tree_node *tnode = subroot->root;
+		/* Take a local copy: array_append_space() below grows the
+		   nodes buffer, which may move it via realloc and leave any
+		   pointer into the old slot dangling. */
+		const struct jwt_node subroot = *array_front(&nodes);
```
sirainen: This `*array_front()` won't compile with non-gcc/clang which don't have enough type checks. Also, I think this could be changed to a one line patch by just adding `subroot = array_front(&nodes);` after the `root = array_append_space(&nodes);`, right?

sirainen (changes requested):

src/lib-oauth2/oauth2-jwt.c
rootvector2: Done. Switched to refreshing subroot after array_append_space() so it survives a possible realloc, instead of the local-copy approach.

src/lib-oauth2/oauth2-jwt.c
```diff
@@ -387,6 +387,7 @@ oauth2_jwt_copy_fields(ARRAY_TYPE(oauth2_field) *fields,
 
 			if (!json_node_is_singular(jnode)) {
 				root = array_append_space(&nodes);
+				subroot = array_front(&nodes);
```
sirainen, line 392: A small comment before this would be good to prevent someone accidentally optimizing away the assignment.

sirainen (changes requested):

src/lib-oauth2/oauth2-jwt.c
rootvector2: Added a comment there.

rootvector2:
any update?

cmouse:
Sorry for delay, this has been moved to our internal system for merging.

cmouse:
merged as 99e93464cff530a8d67e43a49c7cb3516a2fccaa...66d37c58ba05443e03ba28ed66fa4cc8e4f68f87

rootvector2:
thanks, glad this one landed.

PR #291, opened 2026-06-09
OSS-Fuzz: Add new fuzzer targets http request parser

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer targets http request parsing functionality.


sirainen:
I removed the `limits` since they're all defaults. Otherwise merged as f7771362fae54ff78922b8ce420ab043c6a6b903 (will be in github in half an hour)

PR #295, opened 2026-07-11
lib-ssl-iostream: Don't call OPENSSL_cleanup() for openssl 4.0

hector-cao:
With OpenSSL 4.0 the lib-dcrypt unit tests (test-crypto, test-stream) segfault at process exit.

dovecot_openssl_common_global_ref() installs custom OpenSSL memory functions via CRYPTO_set_mem_functions(); their code lives inside the dynamically loaded OpenSSL/dcrypt module. It also passes OPENSSL_INIT_NO_ATEXIT so that libcrypto does not register its own atexit(OPENSSL_cleanup) handler.

dovecot_openssl_common_global_unref() then called OPENSSL_cleanup() explicitly. On OpenSSL 4.0 this explicit cleanup re-arms an internal exit-time cleanup that runs after main() returns [1], i.e. after the module (and therefore the custom memory functions) has been dlclose()d, so libcrypto calls the now-unmapped free() function and the process crashes with SIGSEGV.

[1] https://openssl-library.org/post/2026-03-10-remove-atexit/index.html

### DIFF
--- a/src/lib-ssl-iostream/dovecot-openssl-common.c
+++ b/src/lib-ssl-iostream/dovecot-openssl-common.c
@@ -84,7 +84,7 @@ void dovecot_openssl_common_global_ref(void)
 	   which is after Dovecot has dlclose()d plugins that may have
 	   registered OpenSSL callbacks (e.g. libcassandra).  Accessing those
 	   unmapped code pages from OPENSSL_cleanup() crashes the process.
-	   Dovecot calls OPENSSL_cleanup() explicitly via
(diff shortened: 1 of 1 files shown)


cmouse:
Thanks, we'll take a look.

cmouse:
Hi, your explanation sounds bit counter to the blog post you pointed out, which says that atexit handler is not registered ever.

cmouse:
merged as fda272dc196d368625c425d309ca77241dbf50a1

PR #296, opened 2026-07-13
lib-mail: mbox-from - Fix out-of-bounds read of year field

nishat-06:
mbox_from_parse only checks the minimum date length up front, but the year field is read after a variable number of optional fields (seconds, named or numeric timezone) have been consumed. A From_ line that ends right after those fields leaves msg at msg_end, so mbox_parse_year reads up to four bytes past the buffer. Bound the read against msg_end before parsing the year, matching the msg_end guards used for the other fields. The added test feeds an exact-size heap buffer so the over-read is caught under valgrind/ASAN.

### DIFF
--- a/src/lib-mail/mbox-from.c
+++ b/src/lib-mail/mbox-from.c
@@ -220,7 +220,7 @@ int mbox_from_parse(const unsigned char *msg, size_t size,
 
 	if (!alt_stamp) {
 		/* year */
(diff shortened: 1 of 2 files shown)


nishat-06:
any update?

cmouse:
Merged as f9f517ee574ec2882c7f4aca9739f8c91e40ace4

PR #297, opened 2026-07-14
lib-mail: message-parser - Fix out-of-bounds read on empty preamble

nishat-06:
Re-parsing a multipart message from cached parts with
`MESSAGE_PARSER_FLAG_INCLUDE_MULTIPART_BLOCKS`, on a message whose MIME
part has an empty preamble (the boundary is the first thing in the body):

    block: hdr=(nil) size=18446744073709551614
    Panic: nearest_power(18446744073709551615): calculation would overflow

`preparsed_parse_prologue_more()` scans backwards from the boundary's
trailing CRLF for the newline that precedes `--boundary`. With no preamble
the boundary sits at the very start of the prologue block, so the scan runs
off the front and exits with `cur == block_r->data - 1`. `data[-1]` happens
to be the `\n` ending the multipart part header, so the `--` check passes on
`data[0]`/`data[1]` and `block_r->size = cur - block_r->data` underflows to a
near-`SIZE_MAX` value. The prologue block is then handed to the consumer with
that size, which over-reads (the panic above is a consumer appending it).

Reachable via `message_parser_init_from_parts()` when a caller re-parses a
message against its cached `message_part` tree and asks for multipart blocks.

Handle the no-preamble case explicitly: when the backward scan finds no
newline, confirm the block starts with `--` and return an empty prologue.
Regression test added to test-message-parser.c.

### DIFF
--- a/src/lib-mail/message-parser-from-parts.c
+++ b/src/lib-mail/message-parser-from-parts.c
@@ -121,15 +121,26 @@ static int preparsed_parse_prologue_more(struct message_parser_ctx *ctx,
 			if (*cur == '\n') break;
 		}
 
-		if (cur[0] != '\n' || cur[1] != '-' || cur[2] != '-') {
-			ctx->broken_reason = "Prologue boundary beginning not at expected position";
-			return -1;
+		if (cur < block_r->data) {
+			/* boundary is at the very start of the prologue, i.e.
+			   there is no preamble before it (no preceding newline
+			   to find). Don't read before block_r->data. */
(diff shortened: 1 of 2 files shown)


cmouse:
Thank you, we'll take a look.

nishat-06:
any update?

cmouse:
Hi, thank you for your pull request, we are looking at it.

cmouse:
This is already internally tracked with slightly different solution. We'll see if has any usefulness for that.

PR #298, opened 2026-07-15
lib: uri_parse_ip_literal() - Fix one-byte out-of-bounds read

nishat-06:
uri_parse_ip_literal() scans for the closing ']' with memchr(parser->cur+1, ']', parser->end - parser->cur), but the readable span starting at cur+1 is only end - cur - 1 bytes, so the length is one too large and the scan reads parser->end itself when no ']' is present. It is reachable from uri_check_data()/uri_parser_init_data() with any authority host that starts with '[' and has no ']'; C-string callers happen to land on the NUL terminator, but a non-NUL-terminated (data, size) buffer makes it a real heap over-read. The added test feeds an exact-size buffer so it trips under valgrind/ASAN, and subtracting the extra byte from the memchr length resolves it.

### DIFF
--- a/src/lib/test-uri.c
+++ b/src/lib/test-uri.c
@@ -74,6 +74,26 @@ static void test_uri_invalid(void)
 	test_end();
 }
 
+static void test_uri_ip_literal_overread(void)
+{
+	/* An IP-literal host "[" with no closing "]". Use an exact-size,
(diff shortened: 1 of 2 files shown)


cmouse:
Hi!

Similar fix is already in our internal queue. Thank you for reporting this though.

Aki

PR #300, opened 2026-07-18
doveadm: client-connection-tcp - Don't read past args in flag loop

nishat-06:
client_handle_command_ctx() checks argc>=3 once, before the flag loop, but each 'x' (extra-fields) flag then consumes an args slot with no further check, and the username and command name take two more after it. A doveadm command line whose flag field holds more flag-consumers than there are following arguments makes args walk past the NULL terminator of the t_strsplit_tabescaped() array: with flags "xx" on a three-field line, cmd_name = args[0] reads one entry past the array and argc underflows to UINT_MAX before it reaches doveadm_cmd_handle(). Re-check argc before taking the extra-fields argument and before reading the username and command name.


nishat-06:
any update?

cmouse:
we have most of prs in internal review

cmouse:
Merged as 42fc87ba8f71d63eadb6d764e3546a9b9d943be5

PR #301, opened 2026-07-18
Fix fatal_array test for 32-bit systems

pghmcfc:
On 64-bit systems, `SIZE_MAX` is much bigger than `UINT_MAX` so an allocation of `UINT_MAX` objects is not unreasonable.

On 32-bit systems, `SIZE_MAX` and `UINT_MAX` are the same so an allocation of `UINT_MAX` objects fails `malloc` checks before reaching buffer size limit checks.


cmouse:
Merged as 25a1010e7c36b205f0d51335059ea0e5e2bd4e22

PR #302, opened 2026-07-19
lib-storage: mailbox-list - Track hierarchy part start using ns_sep

nishat-06:
mailbox_list_escape_name_params_to_str() tracks the start of a hierarchy part with a hardcoded '/', so when the caller passes a different ns_sep the "." and ".." checks in need_escape_dirstart() are skipped for every part after the first. imapc passes the remote server's announced LIST separator here along with the remote mailbox name, so a backend using ':' can return "..:..:etc" and get "../../etc" back as the local storage and fs name. Track the part start with ns_sep too, and use that same terminator test inside need_escape_dirstart().

### DIFF
--- a/src/lib-storage/mailbox-list.c
+++ b/src/lib-storage/mailbox-list.c
@@ -191,19 +191,28 @@ const char *mailbox_list_get_unexpanded_path(struct mailbox_list *list,
 	return "";
 }
 
-static bool need_escape_dirstart(const char *vname, const char *maildir_name)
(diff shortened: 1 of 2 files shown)


nishat-06:
any update?

cmouse:
Fixed with https://github.com/dovecot/core/compare/eb28bd214c2dec55be43b040782ce8da3718d3b7%5E...f28b5a92ec52eef0a111a095d00435756343dc94

PR #303, opened 2026-07-29
lib-index: mail-index-sync-ext - Fix offset overflow in ext header update

nishat-06:
mail_index_sync_ext_hdr_update() checks offset + size > ext->hdr_size, but both offset and size are uint32_t read from a transaction log ext-header-update32 record, so the sum wraps for a large offset and the check passes. The buffer_write() at ext->hdr_offset + offset then also wraps and writes the record data outside the extension's own header area in hdr_copy_buf. This checks offset and size separately so the comparison can't overflow, and adds a regression test.


cmouse:
merged as 653b02c7a53ad909e926fb2ccb42a0898ea73e79

PR #308, opened 2026-08-24
auth: db-oauth2 - require all configured scopes in token

nishat-06:
db_oauth2_token_in_scope() treats oauth2_scope as any-of: the loop breaks as soon as the token carries one of the configured scopes, so a token missing the rest still authenticates in introspection and tokeninfo modes. The JWT local-validation path (check_scope in lib-oauth2/oauth2-jwt.c) already requires every configured scope, and test-oauth2-jwt covers that behavior. Fail when any configured scope is absent so both validation paths enforce the setting the same way.


cmouse:
Hi, this is tracked internally as a critical vulnerability. Please contact security@dovecot.org before opening potential security fixes. We can leave this open now, as there is no point to retract this anymore from here.

PR #311, opened 2026-08-27
lib-imap: imap-utf7 - Reject invalid octet in imap_utf7_to_utf8()

nishat-06:
- [ ] I confirm that this PR does not fix, disclose, demonstrate, or discuss a suspected security vulnerability.
- [x] I have read [CONTRIBUTING.md](./CONTRIBUTING.md) and [SECURITY.md](./SECURITY.md)
- [ ] I have compiled and tested this code

## AI policy

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Description

imap_utf7_to_utf8() is supposed to reject invalid mUTF-7, but a raw 8-bit or control octet that appears outside a shift sequence is emitted silently instead: the non-escaped path in imap_utf7_to_utf8_int() (escape_chars is "") falls into the escape branch, runs str_printfa(dest, "%c%02x", '\0', octet) which appends a NUL byte, and still returns success. That disagrees with imap_utf7_is_valid(), which rejects the same input, so a client mailbox name such as `&AAE-\x80` (SELECT/CREATE/LIST via client_find_namespace_full) slips past the "not valid mUTF-7" check and yields a name containing an embedded NUL. The fix returns -1 in the non-escaped path, matching the sibling mbase64 failure branch just below it; the escaped API and every valid input are unchanged.

## Additional Notes

Added test_imap_utf7_raw_octet_after_shift to test-imap-utf7.c. It sweeps every control/8-bit octet placed after a completed shift and asserts both imap_utf7_is_valid() and imap_utf7_to_utf8() reject it. The case fails on the current tree (the octet is accepted with an embedded NUL) and passes with this change, while the existing valid-input and escaped round-trip cases are unaffected.

### DIFF
--- a/src/lib-imap/imap-utf7.c
+++ b/src/lib-imap/imap-utf7.c
@@ -313,6 +313,13 @@ imap_utf7_to_utf8_int(const char *src, const char *escape_chars, string_t *dest)
 	while (*p != '\0') {
 		if (strchr(escape_chars, *p) != NULL ||
 		    *p < 0x20 || *p >= 0x7f) {
+			if (escape_chars[0] == '\0') {
+				/* non-escaped API: an 8-bit or control octet
+				   outside a shift sequence is invalid mUTF-7.
+				   Reject it like the mbase64 failure path below
+				   instead of emitting a NUL escape char. */
+				return -1;
+			}
 			str_printfa(dest, "%c%02x", escape_chars[0],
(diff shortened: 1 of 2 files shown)


cmouse:
Hi! Thank you for the submission, is this a vulnerability change as you have not certified that it is not? Also this look a bit like AI assisted submission.

PR #312, opened 2026-08-31
Remove 'name' attribute from tokenizer elements

beertje44:
both name and class cause a startup error in current solr, it should be only one of those, so only class is retained.

# Pull Request

If this pull request fixes, discloses, demonstrates or discusses a suspected security vulnerability,
**DO NOT submit it publicly**. Follow [SECURITY.md](./SECURITY.MD) and report it privately.

- [X] I confirm that this PR does not fix, disclose, demonstrate, or discuss a suspected security vulnerability.
- [X] I have read [CONTRIBUTING.md](./CONTRIBUTING.md) and [SECURITY.md](./SECURITY.md)
- [X] I have compiled and tested this code'

## AI policy

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Description

Both solr 9.8+ and current 10 versions refuse to load core schema when tokenizer has both name and class properties. Only keeping class here seemed like the right thing to do. After this my dovecot core loads and works withwith solr-10.0.0 on my system.

### DIFF
--- a/doc/solr-schema-9.xml
+++ b/doc/solr-schema-9.xml
@@ -6,7 +6,7 @@
   <fieldType name="string" class="solr.StrField" omitNorms="true" sortMissingLast="true"/>
   <fieldType name="text" class="solr.TextField" autoGeneratePhraseQueries="true" positionIncrementGap="100">
     <analyzer type="index">
-      <tokenizer name="standard" class="solr.StandardTokenizerFactory"/>
+      <tokenizer class="solr.StandardTokenizerFactory"/>
       <filter name="icuFolding" />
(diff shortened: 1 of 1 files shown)


bcrestaniceov:
Confirmed this works on Solr 9.10, but keeping name might be better according to the docs (it's not legacy) - what do you think?

ref. https://solr.apache.org/guide/solr/9_10/indexing-guide/tokenizers.html

beertje44:
Previous edit added class, so that is why I kept that one:

https://github.com/dovecot/core/commit/b10ca9c4cb04d872bdc9160ab9eaf878e41f1eb7

PR #313, opened 2026-09-01
OSS-Fuzz: Add new fuzzer targets imap parsing

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer targets imap parsing functionality.


sirainen:
Merged in https://github.com/dovecot/core/commit/1ebcddce56f02d36e2ca07987bec1009c7cf9fd5 - we also added some improvements on top of it.

PR #318, opened 2026-09-14
lib: buffer - Annotate w_buffer with sized_by(alloc)

LaptopsPlural:
## Summary

`real_buffer.w_buffer` is the private writable allocation whose capacity is tracked by `alloc`. Annotate that relationship with `sized_by(alloc)` when the compiler provides the attribute (no-op otherwise), and in `buffer_alloc()` assign the new `alloc` before publishing the new pointer so the bound remains consistent while the buffer grows.

No change to public `buffer_t` layout, `writable_size`/NUL reservation, or unsafe space helpers.

## AI-assisted disclosure

This change was prepared with AI assistance. I reviewed the diff for correctness and scope before opening this PR.


src/lib/buffer.c
```diff
@@ -12,6 +12,16 @@
    and it's been checked to work correctly with memcpy(). */
 #undef memcpy
 
+/* sized_by(alloc): byte bound for w_buffer. Empty when unsupported so
+   default builds are unchanged. */
+#if defined(__has_attribute) && __has_attribute(sized_by)
+#  define ATTR_SIZED_BY(member) __attribute__((sized_by(member)))
+#elif defined(__has_attribute) && __has_attribute(__sized_by__)
+#  define ATTR_SIZED_BY(member) __attribute__((__sized_by__(member)))
```
sirainen: Probably not useful to check for `__sized_by__`?

sirainen (changes requested):

sirainen:
Should we add this to our malloc() variants?

src/lib/buffer.c
LaptopsPlural: Agreed — dropped the `__sized_by__` spelling check. We now only gate on `__has_attribute(sized_by)` (empty macro otherwise).

LaptopsPlural:
Thanks for the review.

On malloc variants: I think yes as a follow-up, not in this PR. This change is intentionally one concern — private `w_buffer`↔`alloc` plus the capacity-first publish order in `buffer_alloc()` so the attribute stays sound. Spreading `sized_by` onto the pool/`p_malloc` helpers here would widen the surface (and the review) without being required for this buffer.

Happy to open a follow-up that annotates the malloc variants once this lands, if that direction still looks useful to you.

LaptopsPlural:
Thanks again for the malloc-variants question.

I looked at folding that into this tip. The allocator APIs (`i_malloc` / `p_malloc` / `t_*`) are not a `sized_by` fit — that attribute is for ptr+capacity struct members, and those entry points would be a different (`alloc_size`) surface across several headers. The nearest internal companions (e.g. allocfree `pool_block.block`) are interior pointers into the same allocation, which I would rather not annotate here.

Prefer keeping this PR buffer-only (`w_buffer`↔`alloc`, capacity-first publish). Please confirm whether you want a specific malloc-side site folded into this PR, or are fine with a follow-up once this lands. Re-requesting review on the current tip either way.

